PR #2, opened 2018-03-24
Encoded the © symbol with html special entities

nsuchy:
This pull request encodes the © symbol with html special entities.

### DIFF
--- a/views/partials/footer.hbs
+++ b/views/partials/footer.hbs
@@ -1,3 +1,3 @@
 <footer class="footer">
-    <p>© 2018 Pretendo</p>
+    <p>&copy; 2018 Pretendo</p>
 </footer>


ashmonty:
Hey @nathanielsuchy, I'm Ash from the Pretendo Network developer team.

I'm contacting you because you have contributed to our `PretendoNetwork/website` repository in the past, and we're working on relicensing it under the AGPL3 license (https://choosealicense.com/licenses/agpl-3.0/), which protects the code from being used in closed-source projects, while still permitting modification, as long as the source code is made available. To do so, we need to ask every contributor to also relicense their contributions to AGPL3.

Do you agree to relicense your contribution(s) under AGPL3? (Yes/No)

P.S: I'm asking you here because I wasn't sure where to reach you otherwise

ashmonty:
Good point, but also, better safe than sorry. Thanks!

PR #13, opened 2018-11-24
Added more locale stuff, added morgan, removed body-parser, misc changes 

RedDuckss:
Added more entries to the US_en locale. Also added a default "ERROR" locale if the requested one cannot be found (should be changed to a real default). Also changed some internals to fit semantics better and made views use some new locale entries. Added morgan for some route debugging. Removed body-parser since it's built into express now. Made some error returns more clear and modernized some random snippets.


Nybbit:
Just as a readability kind of thing, instead of using `ERROR` for all the placeholder text in [locales/default.json](https://github.com/PretendoNetwork/website/blob/5703e42de224d6c0ba72f0c3cea5b1532ec267f0/locales/default.json), why not use placeholder text that describes that element?

RedDuckss:
I already noted that:

```
(should be changed to a real default)
```

Nybbit:
Ahh, sorry I missed that. I looked over the rest of the PR and it looks good!

PR #15, opened 2018-11-28
Starting bringing over a better comment scheme for methods (see helpers/api.js), addedemail confirmation sending, proper logging, fixed registration

RedDuckss:
Server now sends confirmation emails to the provided email address. The values can not yet be validated, but they are stored so validation can be added at any time. Added Winston for event logging. Winston has a log level system that merges certain log types into the same file. Maybe we should use a custom logger solution? Registration now works (was not working for me), with added reCaptcha protection, password validation, and the storing of usernames/email validation values. Also changed the PID to a number, cuz it's a number


RedDuckss:
Wow I really fucked up the title for this one didn't I

PR #27, opened 2021-04-09
Fixed unclickable buttons & header when small widths

montylion:
The buttons and the header became unclickable as a result of being covered by the `<div class="deco">`.


JipFr:
Looks good to me!

PR #29, opened 2021-04-25
Added Italian localization, minor change to US_en.json

montylion:
Added Italian localization, minor change to US_en.json


RedDuckss:
Looks good to me

PR #31, opened 2021-04-26
Updated Italian localization

montylion:
Updating as discussed in the Discord server

### DIFF
--- a/locales/IT_it.json
+++ b/locales/IT_it.json
@@ -95,7 +95,7 @@
 			},
 			{
 				"name": "Rambo6Glaz",
-				"caption": "Installer della rete e ricerca su console",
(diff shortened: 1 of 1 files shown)


montylion:
I originally made this pull request thinking that the translation I put in the json was "Installer Network", but that's the one in the master branch.
Personally I feel like both "Installer della rete [...]" and "Programma di installazione della rete [...]" are correct, so there isn't really a difference between merging this pull or closing it.

montylion:
In fact, not merging this pull request is probably better, as it would keep the caption to 2 lines, keeping the card the same width and height as the other cards.

Difference between dev and pull

Closing it for now!

PR #41, opened 2021-05-27
Fix broken OpenGraph description tag and update copyright year

montylion:
Came across this while changing some stuff in the localization tester, so I thought I'd fix it.


montylion:
Also updating the copyright year in the footer

PR #61, opened 2021-09-25
Blog content managment system

montylion:
This adds the code for the blog system


public/assets/css/blogpost.css
```diff
@@ -0,0 +1,206 @@
...
+
+.blog-card h1,
+.blog-card h2,
+.blog-card h3,
+.blog-card h4,
+.blog-card h5,
+.blog-card h6,
+.blog-card strong,
+.blog-card a,
+.blog-card a * {
```
JipFr, line 31: Why do you need so many selectors? Aside from the anchor, wouldn't just `.blog-card` be sufficient?

public/assets/css/blogpost.css
```diff
@@ -0,0 +1,206 @@
...
+  width: 100%;
+  border-radius: 4px;
+  border: 1px solid rgba(255, 255, 255, 0.1);
+}
+
+.blog-card iframe {
+  width: 100%;
+  aspect-ratio: 16/9;
+  border-radius: 4px;
+  border: 1px solid rgba(255, 255, 255, 0.1);
```
JipFr, line 101: I think we should make the `rgba(255,255,255,0.1)`s everywhere into custom properties. `--border` will do.

src/routers/blog.js
```diff
@@ -0,0 +1,79 @@
...
+	const locale = util.getLocale(reqLocale.region, reqLocale.language);
+
+	const localeString = reqLocale.toString();
+
+	const fileList = fs.readdirSync('blogposts');
+
+	// We get the info for each blogpost, ignoring the ones starting with _
+	const postList = fileList
+		.filter(filename => !filename.startsWith('_'))
+		.filter(filename => filename.endsWith('.md')) //Ignores other files/folders
```
JipFr, line 23: This is a really small thing but can you do me a favor and put a space after the `//`

src/routers/blog.js
```diff
@@ -0,0 +1,79 @@
...
+
+	// Get the markdown file corresponding to the post
+	let rawPost;
+	try {
+		rawPost = fs.readFileSync(path.join('blogposts', `${postName}.md`), 'utf-8');
+	} catch(err) {
+		logger.error(err);
+		response.sendStatus(404);
+		logger.warn(`HTTP 404 at /blog/${postName}`);
+		return;
```
JipFr, line 63: Put `next()` as well so it just goes to the next router. Better in case we ever add a proper 404 page

JipFr (approved):

PR #66, opened 2021-10-04
Translated missing roles and localization fields.

ninstar:


montylion:
Merging, thanks for your contribution!

PR #76, opened 2021-10-08
Fixed Jip and Jvs swap

MythicalPlayz:
~Why are your names so similar~
Lol


montylion:
Lmao, thanks for your fix

PR #78, opened 2021-10-09
Update KR_ko.json to current website

sjkim04:
As the title says, I have updated the KR_ko.json to match the current website.

PR #79, opened 2021-10-09
Revert string ‘special thanks’ to English

sjkim04:
In #78, @montylion fixed the ‘Special Thanks’ to ‘특별한 감사’.
But, in my opinion, that sounds awkward.
For example, in the UNDERTALE unofficial Korean patch,
the ‘Special Thanks’ string was stayed untouched.

I once again pull request to discuss about this.

### DIFF
--- a/locales/KR_ko.json
+++ b/locales/KR_ko.json
@@ -121,7 +121,7 @@
 		]
 	},
 	"specialThanks": {
-		"title": "특별한 감사",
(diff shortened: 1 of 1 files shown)


montylion:
Oh, I admittedly should have asked about it before changing it, sorry. I thought it was just a mistake that happened when merging the new strings, my bad. Thank you very much for pointing that out!

PR #83, opened 2021-11-22
Remove second "Special Thanks" from Kinnay

Fangal-Airbag:
I dunno if this was intentional but does Kinnay need a second special thanks if they're already in the special thanks section?


JipFr:
Yep, I like having it there :P

PR #90, opened 2021-12-17
Update Japanese and Korean translation to latest translation

sjkim04:
Also added some untranslated sections in the Japanese file.


ashmonty:
Nice, thank you very much!

PR #113, opened 2022-06-20
Stripe start create subscription

jonbarrow:
Adds the ability to create/start a subscription through Stripe


src/routers/account.js
```diff
@@ -494,4 +499,150 @@ router.get('/miieditor', async (request, response) => {
...
+				case 'canceled': // Subscription was cancled
+				case 'unpaid': // User missed too many payments
+					pnid.access_level = 0;
+					break;
+			
+				default:
+					break;
+			}
+
+			await pnid.save();
```
mrjvs, line 640: Events can be received out of order.
Use a timestamp to "ignore" older events.

https://stripe.com/docs/webhooks/best-practices#event-ordering

src/routers/account.js
```diff
@@ -494,4 +499,150 @@ router.get('/miieditor', async (request, response) => {
...
+	} catch (err) {
+		return response.status(400).send(`Webhook Error: ${err.message}`);
+	}
+
+	if (event.type === 'customer.subscription.updated' || event.type === 'customer.subscription.deleted') {
+		const subscription = event.data.object;
+		const product = await stripe.products.retrieve(subscription.plan.product);
+		const customer = await stripe.customers.retrieve(subscription.customer);
+		const pid = Number(customer.metadata.pnid_pid);
+		const pnid = await database.PNID.findOne({ pid });
```
mrjvs, line 623: I recommend using a direct update query instead of saving a requested object. This is prone to race conditions (when you do start doing conditional logic)

src/database.js
```diff
@@ -0,0 +1,25 @@
...
+	accountServerDBConnection = await mongoose.createConnection(accountServerURI, accountServerConfig.options);
+	accountServerDBConnection.on('error', console.error.bind(console, 'Mongoose connection error:'));
+	accountServerDBConnection.on('close', () => {
+		accountServerDBConnection.removeAllListeners();
+	});
+
+	PNID = accountServerDBConnection.model('PNID', PNIDSchema);
+
+	module.exports.PNID = PNID;
+}
```
mrjvs, line 20: This doesn't wait until connected (but you do accept requests while not connected), use the `connect` event to resolve promise, use error event to reject.

example.config.json
```diff
@@ -17,5 +17,20 @@
...
 	},
+	"stripe": {
+		"secret_key": "sk_secret",
+		"webhook_secret": "whsec_secret"
+	},
+	"database": {
+		"account": {
+			"address": "127.0.0.1",
+			"port": 27017,
+			"database_name": "pretendo",
```
mrjvs, line 28: You can technically use the address to force username and password in the generated url. But it's better to allow for inputting them, or simply just have a `connection_string` in the config. (Which is the recommended way by mongo iirc)

src/routers/account.js
```diff
@@ -494,4 +499,150 @@ router.get('/miieditor', async (request, response) => {
...
+
+router.post('/stripe-wh', express.raw({ type: 'application/json' }), async (request, response) => {
+	const stripeSignature = request.headers['stripe-signature'];
+	let event;
+
+	try {
+		event = stripe.webhooks.constructEvent(request.body, stripeSignature, config.stripe.webhook_secret);
+	} catch (err) {
+		return response.status(400).send(`Webhook Error: ${err.message}`);
+	}
```
mrjvs, line 616: On boot, you should use stripe's event list endpoint to recover any potentionally missed data. (Filter on unsuccessful delivery). This will make the service self healing.

https://stripe.com/docs/api/events

mrjvs (changes requested):

src/database.js
mrjvs: Or await the `connection.asPromise()`

src/routers/account.js
jonbarrow: Can you clarify why this is an issue here? Is there a concern with, say, an event with the `active` status coming in after one with `unpaid` for example?

I’m just trying to fully understand the issue before implementing a fix for it

mrjvs: Imagine someone's payment failing, then doing a retry a few minutes after. (Quite common in payment, called soft fails)

That would send two events: unpaid and active. With a few minutes apart.

If the first one fails for any reason (outage, restart, deploy, crash, etc) and the second one succeeds. It will all be fine, the problem starts when stripe does it's retry logic an hour later. It will then send the `unpaid` status and will mark the account as not having paid, even tho they did pay successfully.

By storing the last sync time in the db (gotten from event), you can ignore older events that aren't relevant anymore.

src/routers/account.js
mrjvs, line 623: If account got deleted, this would return undefined and will throw an error later in the code. That returns a 500 and will make stripe infinitely retry this event.

Also if customer doesn't have meta data, this would return first account in db and modify it (equivalent of `findOne({})`). Should probably add a undefined check.

jonbarrow: When you say "account" are you talking about a customer with Stripe or a PNID with us? Because a PNID would never be deleted, only banned at worst (which I will handle later, as I don't want people buying these tiers to essentially unban themselves)

Also a customer should _never_ not have metadata. The customers used here are always created with the metadata. The server checks if a customer can be found with the users metadata before sending them off the checkout, and if it can't find a customer with the metadata it uses that one and forces the checkout to use it. So that webhook will never fire with a customer without metadata

I can always add a check for it anyway, but I can absolutely guarantee it will never happen (besides Stripe having a major fuck up on their end, in which case I might be worried about other things)

Though either way, I'm not really sure what the best course of action would be here? It's not like I can notify the user that the webhook failed, and at this stage the payment has already been processed. So I'm not really sure how to even effectively handle an error like that which would let both us _and_ the user know an issue happened, or how to rectify it?

jonbarrow: Fixed

src/database.js
jonbarrow: Fixed by awaiting `asPromise()`

example.config.json
jonbarrow: Yeah I have it use a full URI in the account server. Not sure why I did it this way here, my bad

jonbarrow: fixed

src/routers/account.js
mrjvs: I wouldn't put it on the customer meta data for two reasons:

1.its a seperate call, then you have two save calls. what is your backup plan if one of them fails?
if either fails, it will result in the account db being wrong in the end (either syncing wrong or being overwritten by an old event)

2. its not related to the customer, its related to the account, not the customer. those are differen entities, even if they refer to the same person.
 
---

Possible solutions:
1. go full microservice mode, have a seperate service with a seperate db for all payment and access level storage.
when any service needs to know the data, they query the microservice.

2. store the timestamp on the account db (easiest, i recommend this one)
sure its not technically related to the account the PNID. but you gotta have some place to store it.

src/routers/account.js
mrjvs: > Because a PNID would never be deleted, only banned at worst
thats fine, but do you really want a webhook to infinite loop just because you removed one test account, or one account was somehow so corrupted that it needed to be removed?

the account db can be controlled from many places, you cant garuantee it will never make one mistake. and thus you should make it so such a small mistake wont cause major failures, just keep a small mistake small by accounting for these things.
its loosly linked data, so it can fuck up.

> Also a customer should never not have metadata
assuming you never add any other payment system. lets say you rent out game servers in the future, that would make this code infinite retry loop again.

again its loosly linked data, easy to get out of sync or used in a different manner. just add a simple check so a simple oversight doesnt become a major issue.

> Though either way, I'm not really sure what the best course of action would be here?
if there is no pnid attached to the customer, just ignore the event and mark as success. its not meant for this service or integration then.
if the PNID account doesn't exist anymore, just ignore the event and mark as success. its probably an event from the past before the account got yeeted.
both should probably log something in the console, just for debugging purposes.

> but I can absolutely guarantee it will never happen
all these cases are if something small accidently happened. its de-escalation of an issue.
an accidental delete shouldn't cause a backlog of failed events. neither should a typo when you were looking through the stripe customer list in the UI.

src/routers/account.js
jonbarrow: > its a seperate call, then you have two save calls. what is your backup plan if one of them fails?
if either fails, it will result in the account db being wrong in the end (either syncing wrong or being overwritten by an old event)

What do you mean separate calls? I'm not pulling the customer data twice, I already pull it once? And I'm not sure what you mean by the `it will result in the account db being wrong in the end` since the entire point of storing it on the customer is to _not_ have it in the account DB? There's nothing there to "be wrong"?

> its not related to the customer, its related to the account, not the customer. those are differen entities, even if they refer to the same person.

I would say it _is_ related to the customer. It's the Stripe customer which is triggering the webhook, not the PNID? So would it not make sense to just store the data relating to webhook the customer triggered on the webhook itself?

> go full microservice mode, have a seperate service with a seperate db for all payment and access level storage.
when any service needs to know the data, they query the microservice.

That is very. Over-engineering imo?

> store the timestamp on the account db (easiest, i recommend this one)
sure its not technically related to the account the PNID. but you gotta have some place to store it.

The fact that it's not related to the PNID is exactly why I don't want to store it there? It's related to the customer

src/routers/account.js
jonbarrow: > the account db can be controlled from many places, you cant garuantee it will never make one mistake. and thus you should make it so such a small mistake wont cause major failures, just keep a small mistake small by accounting for these things.

Calling this a small issue kinda downplays the issue. A PNID being _deleted_ is a major issue by itself. That should never occur, ever, and there's no system in place where that could happen save someone wiping the DB manually?

> assuming you never add any other payment system. lets say you rent out game servers in the future, that would make this code infinite retry loop again.
> 
> again its loosly linked data, easy to get out of sync or used in a different manner. just add a simple check so a simple oversight doesnt become a major issue.

I'm not sure I understand? If I ever rent out game servers like that, it will be done through the PNID. So the same Stripe customer account will end up being used? There isn't a time where checkout would happen with a Stripe customer account that doesn't have a PNID attached, since it will always force one to be made if it can't find one?

> if there is no pnid attached to the customer, just ignore the event and mark as success. its not meant for this service or integration then.
> if the PNID account doesn't exist anymore, just ignore the event and mark as success. its probably an event from the past before the account got yeeted.

the problem with these are that both of these situations should never happen, and should not be allowed. If at any point a subscription webhook fires where the customer account has no PNID, or if the linked PNID can't be found, those are _major_ issues. That's not something small that can be just ignored. Those aren't valid inputs, that's very bad. Custom accounts are are always force created a PNID attached, if one isn't there something major has fucked up. And a PNID should never ever be deleted entirely, if that's the case something major has fucked up

> footnote: if you still disagree, thats fine. im not gonna argue this further.
> just know that you're making it really easy for user-error to be a big issue if you dont add these safety checks

I'm not trying to argue, I'm just trying to make sure we are on the same page with everything here and how the checkout flow goes so that we can reach a reasonable resolution

src/routers/account.js
mrjvs: > What do you mean separate calls?

you have to update the customer meta data with an api call. and you have to update the access level on the account db document. those can both fail.

and `There's nothing there to "be wrong"?`, what can be wrong is the wrong access level on the PNID.
if the timestamp on customer meta-data fails to update then an older event on a retry can insert wrong access level.
if the access level fails to save but the timestamp does update, they will never obtain that acesss level back until next payment flow.

Databases like postgres and mongo have to follow ACID principles. the A of ACID stands for atomicity. meaning that they can save the entire statement or nothing at all. it will not save half the data. which is exactly what you need right now.
you dont have a garuantee for it be atomic if you have two seperate calls between your pieces of data. so you have to store it in one DB.

> That is very. Over-engineering imo?
depends on your architecture. for pretendo, yes it is.

> The fact that it's not related to the PNID is exactly why I don't want to store it there? It's related to the customer
what other option do you have?

jonbarrow:
Alright, seems like that covered everything?

src/util.js
```diff
@@ -68,11 +75,98 @@ function nintendoPasswordHash(password, pid) {
...
+		if (latestWebhookTimestamp && latestWebhookTimestamp > event.created) {
+			// Do nothing, this webhook is older than the latest seen
+			return;
+		}
+
+		if (!pnid && subscription.status !== 'canceled' && subscription.status !== 'unpaid') {
+			// PNID does not exist. Abort and refund!
+			await stripe.subscriptions.del(subscription.id);
+
+			const invoice = await stripe.invoices.retrieve(subscription.latest_invoice);
```
mrjvs, line 116: I'd add a console log here.

src/util.js
```diff
@@ -68,11 +75,98 @@ function nintendoPasswordHash(password, pid) {
...
+
+		if (!customer.metadata.pnid_pid && subscription.status !== 'canceled' && subscription.status !== 'unpaid') {
+			// No PNID PID linked to customer. Abort and refund!
+			await stripe.subscriptions.del(subscription.id);
+
+			const invoice = await stripe.invoices.retrieve(subscription.latest_invoice);
+			await stripe.refunds.create({
+				payment_intent: invoice.payment_intent
+			});
+
```
mrjvs, line 92: I'd add a console log here

src/util.js
```diff
@@ -68,11 +75,98 @@ function nintendoPasswordHash(password, pid) {
...
+		if (!pnid && subscription.status !== 'canceled' && subscription.status !== 'unpaid') {
+			// PNID does not exist. Abort and refund!
+			await stripe.subscriptions.del(subscription.id);
+
+			const invoice = await stripe.invoices.retrieve(subscription.latest_invoice);
+			await stripe.refunds.create({
+				payment_intent: invoice.payment_intent
+			});
+
+			await sendGmail({
```
mrjvs, line 121: Maybe do retry logic. If email is invalid this probably throws on email bounce. Making it infinite loop the events again.
(idk the library tho, idk if it throws on bounce)

(Email is user input, don't let it screw up the system)

So retry like 5 times otherwise give up. (The log from earlier should make sure we see it if it happens)

mrjvs (changes requested):
Improvements look great, just two minor things left.

mrjvs (approved):

PR #117, opened 2022-06-25
Added the ability to change tiers

jonbarrow:


src/routers/account.js
```diff
@@ -616,16 +618,48 @@ router.post('/checkout/:priceId', async (request, response) => {
...
+		if (stripeConnection.price_id === priceId) {
+			response.cookie('error', 'Already subscribed to this tier', { domain: '.pretendo.network' });
+			return response.redirect('/account');
+		}
+
+		try {
+			const subscription = await stripe.subscriptions.retrieve(stripeConnection.subscription_id);
+			stripe.subscriptions.update(stripeConnection.subscription_id, {
+				cancel_at_period_end: false,
+				proration_behavior: 'always_invoice',
```
mrjvs, line 647: this will refund if its a downgrade. is that intentional?

src/routers/account.js
```diff
@@ -616,16 +618,48 @@ router.post('/checkout/:priceId', async (request, response) => {
...
+	}
+
+	const stripeConnection = pnid.get('connections.stripe');
+
+	if (stripeConnection.subscription_id) {
+		// User already has a subscription with us
+		if (stripeConnection.price_id === priceId) {
+			response.cookie('error', 'Already subscribed to this tier', { domain: '.pretendo.network' });
+			return response.redirect('/account');
+		}
```
mrjvs, line 641: this information can be outdated at the time of execution. (webhooks may not have fired yet for example)
use the fetched subscription to do this check.

mrjvs (changes requested):

src/routers/account.js
jonbarrow: No that’s not intentional. I think I may has misunderstood the Stripe docs? What would be the appropriate way to go about this?

Also to be clear do you mean refund or prorate?

mrjvs: if it upgrade from 10 to 20 per month. they will pay the difference for the time.
so if its halfway through the month they will pay 5 for this month. (Because difference is 10 per month. and half of the month would be 0.5 X 10 = 5. the next months it will be 20.

if it downgrades from 20 to 10. they will get refunded the time they didnt use.
so same logic here. they would get refunded 5 dollars.
they didnt use half the money yet, so thats 10 dollars not used yet. and they will pay half the time of the 10 dollars. so 5 dollars they'd need to pay.
5 - 10 = -5. so theyd get refunded 5.

does that answer the question?

jonbarrow: Okay so not a refund then, this prorates. That does answer my question yes

that’s not intentional still, I was under the impression it would just bill immediately with no prorate either way (like how the Patreon was). Do you know what the better option would be?

I was under the impression from the documentation that setting `proration_behavior` to `always_invoice` would bill immediately and not do a prorate

mrjvs: im not sure, you could always cancel subscription and start a new one.
I dont know how to handle this tho, your call.

jonbarrow: Looks like setting `proration_behavior` to `none` disables prorations

mrjvs (changes requested):

src/util.js
```diff
@@ -167,11 +178,19 @@ async function handleStripeEvent(event) {
 				default:
 					break;
 			}
-
-			await database.PNID.updateOne({ pid }, { $set: updateData }, { upsert: true }).exec();
 		}
 
+		await database.PNID.updateOne({ pid }, { $set: updateData }, { upsert: true }).exec();
```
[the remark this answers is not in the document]
jonbarrow: fixed using suggestion from discord

src/util.js
```diff
@@ -139,14 +139,25 @@ async function handleStripeEvent(event) {
 			return;
 		}
 
+		const oldSubscriptionId = pnid.get('connections.stripe.subscription_id');
```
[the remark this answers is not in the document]
jonbarrow: Purely personal choice, plus sometimes I add things on to schemas but forget to update the models so this helps during testing

mrjvs (approved):

ashmonty (commented):
looks good to me

PR #118, opened 2022-07-04
Add tier change styling

ashmonty:


oscie57:
imagine needing style, plain html is good enough @ashmonty 👍🏻

src/routers/account.js
```diff
@@ -524,7 +524,8 @@ router.get('/upgrade', async (request, response) => {
 		layout: 'main',
 		locale: util.getLocale(request.locale.region, request.locale.language),
 		localeString: request.locale.toString(),
-		error: request.cookies.error
+		error: request.cookies.error,
+		currentTier: 'price_1LBnZADOJlJAaQQ3pEUjNWbY' // To be replaced
```
mrjvs, line 528: cant this be implemented now that the other PR was merged?

mrjvs (changes requested):

src/routers/account.js
ashmonty: It probably can. psst @jonbarrow

jonbarrow: @ashmonty @mrjvs I will add that in with my next feature PR, I still need to add the cancellation feature

ashmonty: Okie-dokie!

mrjvs (approved):

ashmonty:
> imagine needing style, plain html is good enough @ashmonty 👍🏻 

the thing is I've got plenty style, and keeping it all to myself would be kinda greedy 😎

PR #123, opened 2022-07-16
Large code cleanup

jonbarrow:
This PR makes many changes to make the backend codebase a bit easier to work with. These changes include:

- Moved duplicate renderData code to middleware
- Moved redirects to middleware
- Moved duplicate auth checking and account grabbing to middleware
- Moved API calls to `util.js` with refresh and retries on expired tokens
- Improved Discord connections and data requests
- Users now need to input their password to download Cemu files (removing the need to store their password hash as a cookie after login)
- Many smaller bug fixes/minor improvements


public/assets/js/account.js
```diff
@@ -5,7 +5,7 @@ document.getElementById('remove-discord-connection')?.addEventListener('click',
 	const tokenType = document.cookie.split('; ').find(row => row.startsWith('token_type=')).split('=')[1];
 	const accessToken = document.cookie.split('; ').find(row => row.startsWith('access_token=')).split('=')[1];
 
-	fetch('https://api.pretendo.cc/v1/connections/remove/discord', {
+	fetch('http://api.pretendo.cc/v1/connections/remove/discord', {
```
mrjvs, line 8: why?

mrjvs (changes requested):

public/assets/js/account.js
jonbarrow: Was having issues with my local machine, can be changed back

public/assets/js/account.js
```diff
@@ -19,4 +19,44 @@ document.getElementById('remove-discord-connection')?.addEventListener('click',
...
+		headers: {
+			'Accept': 'application/json',
+			'Content-Type': 'application/json'
+		},
+		body: JSON.stringify({ 
+			password: onlineFilesModalPasswordInput.value
+		})
+	})
+		.then(response => response.blob())
+		.then()
```
ashmonty, line 47: Is this stray await supposed to be here?

ashmonty (changes requested):
After a quick glance I think this is all?

public/assets/js/account.js
ashmonty: welp, that selected more lines it should have. it's 47

jonbarrow: Fixed

ashmonty (approved):

PR #124, opened 2022-07-23
css cleanup

ashmonty:
Now adhering to this palette

palette

### DIFF
--- a/public/assets/css/account.css
+++ b/public/assets/css/account.css
@@ -12,7 +12,7 @@ body.modal-open {
   display: grid;
   column-gap: 48px;
   margin-top: 80px;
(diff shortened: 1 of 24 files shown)


jonbarrow:
Should the components that we end up reusing (like buttons, modals, etc) be moved to (a) dedicated file(s)? That way they aren't redefined multiple times in the styles?

ashmonty:
> Should the components that we end up reusing (like buttons, modals, etc) be moved to (a) dedicated file(s)? That way they aren't redefined multiple times in the styles?

yeah, can do!

ashmonty:
did done do!

PR #131, opened 2022-07-30
feature: mii editor

ashmonty:
Add icons, mobile layout and save animation to mii editor
Add styling for range input element to components.css
Clean up account.js route by moving editor json to separate file


ashmonty:
hell yeah

PR #135, opened 2022-08-13
feature: miieditor

ashmonty:


jonbarrow:
There's no button to go to the Mii editor from the account page (maybe clicking the Mii icon?)

Also when changing the Mii's body size the image moves around instead of being locked in one place. You can see this when changing the Mii's height, set it to 0 and then move it all the way to max

jonbarrow (changes requested):
There's no button to go to the Mii editor from the account page (maybe clicking the Mii icon?)

Also when changing the Mii's body size the image moves around instead of being locked in one place. You can see this when changing the Mii's height, set it to 0 and then move it all the way to max

ashmonty:
> There's no button to go to the Mii editor from the account page (maybe clicking the Mii icon?)

Yeah, I was also thinking of clicking the Mii icon. Will add!

> Also when changing the Mii's body size the image moves around instead of being locked in one place. You can see this when changing the Mii's height, set it to 0 and then move it all the way to max

Ok I swear we've talked about this before. It's just a quirk with the Mii renderer. While it may look like the Mii is moving around, when the image loads if you check the position of the mii's feet, and the width of the mii's head, they should be consistent across any height. If you have a better solution for this, feel free to suggest, but I'm pretty sure the alternative would be to have the Mii be the same height (making the actual Mii height unclear) and/or having it move around and *not* be consistently positioned after the image load.

jonbarrow:
> Ok I swear we've talked about this before. It's just a quirk with the Mii renderer. While it may look like the Mii is moving around, when the image loads if you check the position of the mii's feet, and the width of the mii's head, they should be consistent across any height. If you have a better solution for this, feel free to suggest, but I'm pretty sure the alternative would be to have the Mii be the same height (making the actual Mii height unclear) and/or having it move around and not be consistently positioned after the image load.

We definitely need to find a better solution to this. Because [as this video shows](https://embed.so/video/4926ce7b-f84f-4e7e-81b2-dc31698cd605) the Mii jumps around and rapidly changes size creating a less-than-amazing user experience

Nintendo's Mii Studio does not have this issue, [as seen here](https://embed.so/video/e48271ef-228c-4a03-9f7a-6c58deea5d73), likely due to the fact that they render the Mii on a canvas element instead of actually displaying a raw image

ashmonty:
Yeah I'm aware Nintendo's Mii Studio doesn't have that issue, as, as you mentioned, they use a canvas (which is not something we can do). Even if I went back to only moving the shadow, not the mii itself, the mii would still shift because that's just how the api gives the mii back. I'll think about it, but I'm not sure there's really a solution for this.

ashmonty:
Also here's the link
image image

jonbarrow:
Maybe @mrjvs has some ideas? I'm half tempted at this point to check out Mii Studio's JS and see how their canvas renderer works

mrjvs:
So the problem seems to be that the image doesn't load fast enough, we can't really spam the rendering api that much either.

Having an animation of sorts to make it a less jarring jump will definitely help. (With denounce for slider)
It'll also help to have the feet consistently on the same spot (should be same curve as the shadow, just applied to the image instead of shadow)

I dont think the answer is to poke around in the miistudio js code, cuz its just a rendered and that's precisely what we don't want to recreate.

ashmonty:
> So the problem seems to be that the image doesn't load fast enough, we can't really spam the rendering api that much either.
Yeah

> Having an animation of sorts to make it a less jarring jump will definitely help. (With denounce for slider) It'll also help to have the feet consistently on the same spot (should be same curve as the shadow, just applied to the image instead of shadow)

The thing is it already does that. It's just that when the image loads, the mii is in a different place *in the image itself*, so there's no real solution here. I am already applying the curve and all.

> I dont think the answer is to poke around in the miistudio js code, cuz its just a renderer and that's precisely what we don't want to recreate.
Agree

ashmonty:
If you'll pardon the quality, as you can see [here](https://user-images.githubusercontent.com/49426949/184530345-44948b72-d168-4198-b756-243febf6776f.mp4) the mii's feet are approximately in the same place for every height

ashmonty:
fixed!

PR #142, opened 2022-09-01
Mention RiiConnect24 in the Wii FAQ

osfanbuff63:
Some of the Wii's custom servers are hosted by @RiiConnect24 not Wiimmfi (they do different things, but are both worth mentioning here), so this adds a link to rc24.xyz, RiiConnect24's website.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,17 @@
-# Website for Pretendo Network
+# Website
 
-This is the Repository for the Website of the Pretendo Network.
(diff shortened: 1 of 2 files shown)


jonbarrow:
RC24's goals are not inherently the same as Pretendo's. Wiimmfi is the closest analogue to Pretendo for the Wii, which is why only it was mentioned and our decision to recommend only Wiimmfi was based on the general developer consensus

PR #145, opened 2022-09-18
Made it more clear which account features work currently

jonbarrow:
These changes make it more clear which features on the account dashboard actually do and do not work at the moment. Ideally these things should be working however the move to gRPC is not finalized and these features can wait


ashmonty (changes requested):
Jon I will murder you. Add the strings to the locale.

PR #146, opened 2022-09-18
Added docs and donate links to navbar

jonbarrow:


ashmonty (changes requested):
Jon I will murder you. Add the strings to the locale.

PR #147, opened 2022-09-18
Fixed locale not displaying on docs welcome page

jonbarrow:
Docs welcome page would display [object Object] for all localized strings due to the locale helper not taking in the extra prop and the locale being passed incorrectly in the view


ashmonty:
Whoops, messed this one up

PR #149, opened 2022-09-18
Feature assign discord roles

jonbarrow:
This PR adds the ability for the website to assign and remove Discord roles whenever a user purchases a tier, removes their tier, links their Discord account, and removes their Discord account


ashmonty:
How does this deal with the Tester role (which is currently being used to give access to people without having them donate)? Will @Testers still have access to the servers?

jonbarrow:
The Discord role has no affect on access to the game servers, only access to the Discord channels. When linking a Discord account it only adds roles not removes them. When unlinking a Discord account, it checks if you have a purchased tier first and then removes the roles. If a user has no purchased their then the roles stay, so official testers will still have their roles

ashmonty:
Oh, alright.

PR #158, opened 2022-09-20
hotfixes

mrjvs:
check #157 for changes


mrjvs:
approved verbally by jon

PR #169, opened 2022-09-23
Hotfix miieditor

### LINKED ISSUE #154: mii editor failing to save or load
few issues:
 - [ ] new accounts get invalid mii data
 - [ ] mii editor does a `http://` request instead of `//` so it cannot save
 - [x] api.pretendo.cc does not have a valid certificate, so will fail to save

to fix number 1, I recommend a modal that says something like "detected broken miidata [make new MII] [go back]"

---

encoded mii data from new accounts that doesnt work:
```
AwAAQOlVognnx0GC2qjhdwOzuI0n2QAAAGBqAHYAcwBqAHYAcwAyAAAAAAAAAEBAAAAhAQJoRBgmNEYUgRIXaA0AACkAUkhQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAHwu
```

mrjvs:
resolves #154 

- uses https for saving
- if it detects broken mii data, it asks if they want to start a new mii instead

note: new mii name is steve

### DIFF
--- a/public/assets/js/miieditor.js
+++ b/public/assets/js/miieditor.js
@@ -6,6 +6,7 @@
  * browserify is needed for the use of require() in the browser
  */
 const Mii = require('mii-js');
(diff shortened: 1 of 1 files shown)


ashmonty (approved):

ashmonty:
sbeve

PR #172, opened 2022-09-24
reformatted-mii-icon-account

CirnoMoment:
Added a margin to the mii icon on the account page to fix off-centering


SmellingRoom3:
erm... WHY ?

CirnoMoment:
Why? It's because the mii icon in the account page is off centered, as seen here:
image

ashmonty:
Thank you for your contribution! However, this was already fixed in 0e3799983e5deb43a2c6ca34eafd1c46ba7ee75d

PR #173, opened 2022-09-24
fixed-scrolling-past-footer

CirnoMoment:
Now you shouldn't be able to scroll past the footer due to the size of the oval.


ashmonty:
LGTM! @mrjvs any thoughts?

mrjvs (approved):

PR #174, opened 2022-09-25
Update en_US.json

oscie57:


oscie57:
I do not own ANY content shown. This includes, but is not limited to, music and background images. I do not monitize videos, I do not make profit from videos. I promote artists. If any copyright owner has an issue with my posting of their content, please email halolink44@gmail.com about the issue. I will take down the video immediately.

ashmonty:
I don't know, I feel like this almost has a semblance of truth to it... I wouldn't want people to get confused and think this is actually something we do

oscie57:
oh true, couldnt you add a disclaimer or sm?

ashmonty:
It's just the praxis idk


all your base are belong to us

PR #177, opened 2022-10-01
[Feature] Password resetting

jonbarrow:
Adds the ability to reset passwords from the website. The associated account server changes have already been pushed to the `dev` branch of the account server repo.

Desperately needs frontend work done. I slapped together some frontend pages but they use way too much CSS and the notifications are standard `alert` calls


PokeyManatee4:
Nice

PR #178, opened 2022-10-03
a11y improvements for images and buttons in home page

Jeith:
Updated all images and svgs to have alt text. Used alt="" to mark [decorative images](https://www.w3.org/WAI/tutorials/images/decorative/), and included aria labels for buttons & links without text.


ashmonty (approved):
Looks good to me! @jonbarrow any thoughts?

jonbarrow:
Looks good to me!

PR #181, opened 2022-11-05
Documentation "overhaul"

jbmagination:
**[This pull request is not ready to be merged. Feedback is still appreciated.]**

This pull request makes major changes to the documentation.

A current list of changes are as follows:
* Recently, instructions on how to make a PNID were added. This goes a bit more in-depth on that
  * I find many still ask in #support how to make one, or why it continues to mention Nintendo Network
  * Hopefully mentioning specifics can fix that
* Grammatical issues were fixed
* Going directly to a download is prioritized over navigating through a GitHub Releases page
  * While I have not seen it specifically cause confusion amongst Pretendo users, I have seen it cause confusion elsewhere
* The "Legacy" section on Wii U has been changed to "Other"
* Removing `50_hbl_launcher.rpx` when installing Nimble has been mentioned
  * This was another thing asked about a lot in #support, and it actually took me a while to figure out
* Mentioning Luma patches has also been added
  * Luma patches were something already mentioned in the documentation, but there was nothing asking to ensure they were actually enabled
  * This is now the case
* Juxt error codes have been removed from the sidebar
* In `/docs/install`, the "3DS" option label has been changed to say "3DS/2DS Family"
* Some questions in the FAQ have had their answers updated
  * This was done in English - other languages need to be updated accordingly
* Many images have been changed for consistency
  * The Pretendo Network ID (`PretendoDocs`) was used (I would say that I'm happy to hand that ID over to the team, but I'm also sure you don't need my permission to do that)
* On the bleeding edge installation instructions, logging in on GitHub has been mentioned
  * This confused me at first when trying to follow along until I realized I was dumb lmfao

Changes that still need to be made before this can be marked as ready:
* 3DS PNID linking instructions
* Uninstallation instructions
* Whatever other feedback people provide


jonbarrow:
Just glanced over this, will look more in depth later today, but looks good so far!

Can I ask why there were some changes to the existing images though? For example the SD card images as well as the images showing the `Actions` tab on the repos?

jbmagination:
The images were changed for consistency; the Actions tab image also included extra UI elements from what appears to be other extensions so I redid it to not include those

jonbarrow:
That's fair, I use linux while @CaramelKat uses MacOS

However you did remove a number of images without any replacements?

For example you removed the following files:

- `public/assets/images/docs/install/3ds/releases-highlight.png`
- `public/assets/images/docs/install/3ds/zip-highlight.png`
- `public/assets/images/docs/install/juxt/martini-sd-card.png`
- `public/assets/images/docs/install/wiiu/aroma/releases-highlight.png`
- `public/assets/images/docs/install/wiiu/aroma/wps-highlight.png`
- `public/assets/images/docs/install/wiiu/tiramisu/releases-highlight.png`
- `public/assets/images/docs/install/wiiu/tiramisu/rpx-highlight.png`

But did not replace them with anything more consistent that I can tell?

Though again I am just looking at the changed files tab here, I have not pulled your changes down or ran anything yet

Do you have a live version to check?

jbmagination:
Some images were removed because they're no longer referenced in the docs, or have been moved elsewhere.

Yes, I have a live version - I have been using GitHub Codespaces to make sure that my changes are working as expected.

jonbarrow:
Can I be linked to the live version please? So I can see how it looks

jbmagination:
Will locally set up - you can't access a port on a Codespace unless you're logged in to the account that owns it

jbmagination:
I feel extremely dumb - I have now changed the port visibility to public and you can see the changes live https://jbmagination-improved-space-eureka-54v5v6qwpp2jjp-3000.preview.app.github.dev/docs/welcome

jonbarrow:
Thanks! Few things:

### On the WiiU docs

- The `Already have Pretendo installed, but need help making your Pretendo Network ID?` uses a broken link
- I'm not sure the entire "How to make a PNID" section should be there, at least not the instructions for making it on a console. The console already tells you how to make an account, there's no real reason to say it again?
- All the bits about passwords not being able to be reset and emails not being sent out should be removed. Password resetting and email verification will be merged in before the dev branch gets merged into master and deployed again. There are already PRs for these features
- The `102-2402` on startup only happens if you have a user set to auto login to, and that user is a NNID. Otherwise you wont see that error at all
- I dont exactly agree with removing the Juxt category and moving the instructions to the console categories. Many people will be looking for the Juxt instructions specifically. They were their own category for a reason
- The Martini link uses a hard-coded v0.3 version instead of the latest version
- Does the YouTube app really need it's own dedicated section? No other title has this section, it would probably be better to not have it there and just use the regular progress page (which is something I can do later)

### On the 3DS docs

- I'm nervous about using your SD card screenshot as it features FBI, which is a known tool used for piracy. I'd like to not have any possible connections with us to piracy
- Same thing about creating a PNID and the Juxt docs being moved

### Other

This isn't an issue with your docs but since we are here: If you REALLY want to keep the "How to make a PNID" docs I would suggest simplifying them (again the consoles already tell you what to do, no need to repeat it) and making it it's own section

Also the error code search seems a little odd to me, since real error codes dont use the `ABC-` prefixes. Seems confusing to me, I would suggest just having the error codes visible and not the modules. Maybe we should also remove the Juxt error thing on the sidebar? Seems out of place given that it only lists a few errors and no other module has this support

And one last thing: On pages like the Wii U docs where theres several sections (Tira, Aroma, etc) maybe some way to more easily see the split would be nice? Like on the error code pages

jbmagination:
Thanks!

* The broken link should be fixed now
* The reason I mention doing it on console is because some people still ask in #support where to go. I will see what I can simplify about explaining that process
* Mentions of a lack of email verification and password resets have now both been removed
* I have actually also seen `102-2402` when modifying my user. I doubt that most will ever experience that, but that's why I say it's only usually a good thing - in extremely rare cases it may not be
* The reason I chose to move Juxt into each console's page is a) most will actually be looking to interact like they did on Miiverse, and the only way they can do that is by setting up a Pretendo Network ID, and b) as such, it feels awkward having two separate pages to get people to go to whenever they specifically want Juxt / Miiverse. If we really do want it to be two separate pages, I think we need to make it clear that the only way of interacting with Juxt is by setting up Pretendo Network on their device, and link back to the Juxt setup page at the end of the Wii U and 3DS setup guides
* The Martini link being hardcoded is because there is actually no "latest" version that GitHub has - they're all marked as pre-releases, so using a latest link 404s. If that changes then I can update it to be a latest link
* YouTube specifically doesn't need to be mentioned, no, but [it appears that people are learning about Pretendo solely because of the YouTube patch](https://www.reddit.com/r/wiiu/comments/yhf060/comment/iudydmm/) so I think it'd be nice to accommodate for that in the documentation somewhere. I'd expect the same for other patches that may come in the future, although it might be better to just put the YouTube patch directly into Inkay (and/or Nimble)
* The 3DS docs including FBI is because [that's what 3ds.hacks.guide has you install](https://3ds.hacks.guide/finalizing-setup). I've never heard of FBI being known for piracy, but that's interesting
* I find the error code search strange too, as I do with the Juxt errors on the sidebar, although I'm not sure what to do with the search. I'll remove the sidebar errors, and as long as others also agree to remove the module prefix then I imagine that can be done without my help

Since you're here, I should ask - `102-2534` appears when an endpoint doesn't exist yet, correct?

jonbarrow:
(Since some things have been resolved and do not need to be address, I will be leaving comments using a numbered list. Some numbers will be missing since I'm skipping to other bullet points you made)

2. Simplifying to something like "Create an account on your console like you would any other NNID" should be enough. There's no reason to duplicate the entire account login/registration process when the console does it for you already. If people are asking where to go we simply need to point them in that direction
4. `102-2402` is `BAD_FORMAT_REQUEST` in the `ACT` module. This happens whenever a bad request happens, it should not exactly be called "good" as it's misleading and be thrown for any number of reasons. You likely experienced it while trying to modify your user since changing user settings on the console isn't supported yet, which is a real error not a good thing
5. Pretendo can be used completely fine without Miiverse/Juxt being setup, and I know of several users who do just that (myself included, I've never once actually even opened Miiverse on my Wii U). Since Juxt is not required for the rest of the network to function I think it would be best to keep it as it's own section as an optional setup step. I do agree with making it clear that you NEED a PNID to use Juxt however. Maybe instead of putting the Juxt instructions on each consoles page, simply say somewhere on the page "If you would like to optionally setup Miiverse, go here" with a link. On the Juxt docs add a warning to the top alerting people they need a PNID setup to use it
6. Oh okay I wasn't aware there was no actual release, sounds good
7. We have already started talking internally about moving the YouTube patch to Inkay/Nimble, yes. I still do not think it really needs it's own dedicated section, as if we keep adding dedicated sections for every single patch that will get messy very quickly
8. Oh alright! I was not aware it had you install it in that guide, and I was unaware it was in the original screenshot as well (I didn't write those docs). In my experience I have _only_ ever seen FBI used for piracy, especially using it's QR code installer on "that piracy site", hence my worry. If the official guide has it then it should be fine, but maybe for cleanliness reasons the screenshots should only include the files required for us? That way people don't think they need FBI to use Pretendo, because you don't
9. Sounds good, I'll send someone over to also review that idea shortly
10. `102-2534` is `REQUEST_NOT_FOUND` in the `ACT` module, so yes

jonbarrow:
Also the live version is down again

jbmagination:
The live version should now be back up

jonbarrow:
https://jbmagination-improved-space-eureka-54v5v6qwpp2jjp-3000.preview.app.github.dev/docs/welcome is giving 404

jbmagination:
Sorry for the wait! While working on this last month, [GitHub Codespaces went out of beta, and subsequently, I lost access to Codespaces for the rest of the month](https://github.blog/changelog/2022-11-09-codespaces-individual-beta-end/). Now that I can access the Codespace again, I can get back to working on this (and in a few hours, I will put up a live version on [ngrok](https://ngrok.com/))

docs/en_US/install/cemu.md
```diff
@@ -1,36 +1,42 @@
...
+	Only experimental builds of Cemu 2.0 are supported. At this time, Cemu does not have a stable release of Cemu 2.0 which supports Pretendo.
 </div>
 
-Cemu 2.0 has official built-in support for Pretendo as of October 10, 2022. Head to the Cemu GitHub [releases](https://github.com/cemu-project/Cemu/releases) page and download the latest Cemu experimental release (tagged as `Pre-release`). Only `Cemu 2.0-5 (Experimental)` and above is supported at the moment. Additionally you may build Cemu from source using the provided [build instructions](https://github.com/cemu-project/Cemu/blob/main/BUILD.md)
+Cemu 2.0 introduced official built-in support for Pretendo on October 10th, 2022. Head to the Cemu GitHub [releases](https://github.com/cemu-project/Cemu/releases) page and download the latest Cemu experimental release (tagged as `Pre-release`). Only `Cemu 2.0-5 (Experimental)` and above is supported at the moment. Alternatively, you may build Cemu from source using the provided [build instructions](https://github.com/cemu-project/Cemu/blob/main/BUILD.md).
```
MythicalPlayz, line 11: I think it should be bumed up to 2.0-16 as there were some pretendo bugs I fixed (mainly content-type not being set correctly buy)

PR #186, opened 2022-11-07
Small website bugfixes

mrjvs:
changes:
 1. added `components.css` to the head of blogposts. without it the navbar has unstyled content.
 2. fixed footer logo, it used to be invisible on firefox.
 3. fixed rare FOUC by adding a section helper to handlebars. (it teleports content from partials to the head or end of body)

I was not able to test every page I changed, so I cannot garuantee every page works.
if someone can verify, that would be great.

### DIFF
--- a/public/assets/css/partials/footer.css
+++ b/public/assets/css/partials/footer.css
@@ -25,7 +25,6 @@ footer div {
 }
 footer svg.logotype {
 	height: 56px;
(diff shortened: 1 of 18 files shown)


jonbarrow:
Looks good to me! What do you think @ashmonty?

ashmonty (approved):
looks good to me too!

PR #187, opened 2022-11-19
Account Page Improvements

MythicalPlayz:


sladewatkins:
Not involved with this patch, but nevertheless, this two commits get my ack.
Acked-by: @sladewatkins

src/routes/account.js
```diff
@@ -112,8 +112,8 @@ router.get('/register', async (request, response) => {
 });
 
 router.post('/register', async (request, response) => {
-	const { email, username, mii_name, password, password_confirm, 'h-captcha-response': hCaptchaResponse } = request.body;
-
+	const { email, mii_name, password, password_confirm, 'h-captcha-response': hCaptchaResponse } = request.body;
+	const username = request.body.username.toLowerCase()
```
gitlimes, line 116: Isn't this going to lowercase the actual username? Like, if a user chooses PN_jon, it's going to become pn_jon?

gitlimes (changes requested):

src/routes/account.js
MythicalPlayz: That one I do remember tho, there was the bug when signing up through the websites tht it would reject if the username has a Capital letter(s), as PNIDs are not case-sensitive I added this to hopefully fix the bug.

gitlimes: yeah, but, again, would it not do what I said?

PR #191, opened 2022-12-24
fr-FR translation : Make "remerciement" (thanks) plural

MininuxDev:
It's better to put it in plural, there are multiple people to thank

EDIT: my bad i just realised there is now a weblate for this, should i delete the pull request or leave it ?


MarisaGaming01:
If you want to translate, I’d suggest checking out `#localization` on the Pretendo Discord Server and going to Weblate, where we do translations now.

PR #210, opened 2023-04-08
Add Martini errors, update Juxt docs for new Inkay build

ashquarky:
Gearing up for the applet-patch enabled Inkay, which adds a new error code `598-0069`
Also adds all the legacy Martini errors so people can uninstall and stuff


ashquarky:
Possible issues:
- Linking to wiiu.hacks.guide
- "..Miiverse applet, reinstall a clean copy from NUS or disc. Instructions for this cannot be provided by Pretendo."
- ⚠️ UTF-8 emoji (some kind of infobox like Wikipedia's ones would be nice there)
- 678-1010 and 678-1011's handling as brick risks

docs/en_US/errors/MRTI-678-1005.md
```diff
@@ -0,0 +1,24 @@
...
+
+***Note:*** This error relates to the Martini patches, which are now deprecated.
+
+The patcher could not find an unmodified copy of the Miiverse applet. Either you have manually modified it, it has
+become corrupted, or you have a version too old or too new for the patcher to recognise.
+
+- If you are on a system version older than 5.1.0, perform a system update, install
+  [Tiramisu or Aroma](https://wiiu.hacks.guide), and try again.
+- Try disabling any content replacement plugins (SDCafiine etc.) that can change which files are read by the patcher.
+- If you have manually modified your Miiverse applet, reinstall a clean copy from NUS or disc. Instructions for this
```
jonbarrow, line 14: You know I get hyper worried about things, and wording scares me

Are you against saying something like "reinstall from a backup" instead of directly mentioning NUS here? Saying just "backup" is ambiguous enough to where it could mean either a user made backup of their own copy, or. You know. How pirates refer to "backups". That way we can at least argue we didn't push people in the "go look on NUS" direction?

(only gonna comment this one time despite it happening twice)

docs/en_US/install/juxt.md
```diff
@@ -17,7 +17,10 @@ Nimble and Inkay already take care of in game patches for miiverse, so you're al
 
 <br>
 
-### Applet Patch
+### Applet Patch (Tiramisu)
+**Aroma users**: The Aroma patcher already handles the applet patch and redirects the Miiverse applet to Juxtaposition.
+The below instructions are **only for Tiramisu users.**
+
```
jonbarrow, line 23: You mentioned wanting an infobox, I assume you meant for here. We do have an implementation of an infobox already, seen here https://github.com/PretendoNetwork/website/blob/master/docs/en_US/install/3ds.md?plain=1#L3-L5

docs/en_US/errors/MRTI-678-1005.md
ashquarky: you know how I feel about system titles and the NUS, but I'll defer to you on this one.

I wrote a little bit on the fence about the "backup" phrasing (since Martini is an app that creates said backups) but then I remembered that this sentence starts with "If you have manually modified your Miiverse applet" so I think I'm cool with that change

docs/en_US/install/juxt.md
ashquarky: Is there a warning infobox too? Or just `tip` for now

jonbarrow: Just tip right now, but a warning can easily be added. The emoji can be changed to the warning emoji for now too

This is what the infobox looks like when rendered right now Screenshot from 2023-04-11 07-54-33

ashquarky: I'll just change the emoji I think

docs/en_US/errors/MRTI-678-1005.md
jonbarrow: I know how you feel, I'm not trying to discount or invalidate that, you know I try to just keep everything here in the up-and-up whenever I can

Sounds good, then

PR #218, opened 2023-07-25
removes discriminator for pomelo usernames

oscie57:
quick fix, first time attempting anything with handlebars so it can probably be optimised, though it works for now (i think)

### DIFF
--- a/views/account/account.handlebars
+++ b/views/account/account.handlebars
@@ -145,7 +145,7 @@
 				<div class="setting-card">
 					<h2 class="header">{{ locale.account.settings.settingCards.discord }}</h2>
 						{{#if discordUser}}
-					<p>{{ locale.account.settings.settingCards.connectedToDiscord }} {{ discordUser.username }}#{{ discordUser.discriminator }}</p>
+					<p>{{ locale.account.settings.settingCards.connectedToDiscord }} {{ discordUser.username }}{{#ifCond discordUser.discriminator '#0'}}{{else}}#{{ discordUser.discriminator }}{{/ifCond}}</p>
 					<a href="/account/remove/discord">
 						<button class="button secondary" id="remove-discord-connection">{{ locale.account.settings.settingCards.removeDiscord }}</button>
 					</a>
--- a/views/account/account.handlebars
+++ b/views/account/account.handlebars
@@ -145,7 +145,10 @@
 				<div class="setting-card">
 					<h2 class="header">{{ locale.account.settings.settingCards.discord }}</h2>
 						{{#if discordUser}}
-					<p>{{ locale.account.settings.settingCards.connectedToDiscord }} {{ discordUser.username }}{{#ifCond discordUser.discriminator '#0'}}{{else}}#{{ discordUser.discriminator }}{{/ifCond}}</p>
+					<p>{{ locale.account.settings.settingCards.connectedToDiscord }} {{ discordUser.username }}
+					{{#if discordUser.discriminator}}
+					&nbsp;#{{ discordUser.discriminator }}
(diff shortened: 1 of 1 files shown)


oscie57:
idk why that random pull request merge is there

views/account/account.handlebars
```diff
@@ -145,7 +145,7 @@
 				<div class="setting-card">
 					<h2 class="header">{{ locale.account.settings.settingCards.discord }}</h2>
 						{{#if discordUser}}
-					<p>{{ locale.account.settings.settingCards.connectedToDiscord }} {{ discordUser.username }}#{{ discordUser.discriminator }}</p>
+					<p>{{ locale.account.settings.settingCards.connectedToDiscord }} {{ discordUser.username }}{{#ifCond discordUser.discriminator '#0'}}{{else}}#{{ discordUser.discriminator }}{{/ifCond}}</p>
```
jonbarrow, line 148: - There is no handlebars function called `ifCond`. This crashes
- Even if this didn't crash, `discordUser.discriminator` is a number, it does not contain the `#`, which is why it is added here. These checks would always fail since you seem to be checking for `#0` and not `0`
- We prefer to use full blocks for if statements instead of single lines. See the `#if` checks earlier in this same view for checking the users `environment` status

jonbarrow (changes requested):
Please test your code before making PRs

views/account/account.handlebars
oscie57: should be fixed, however i was unable to rehost the website as i was missing critical files for the account server and didnt know how to get them

views/account/account.handlebars
```diff
@@ -145,7 +145,10 @@
 				<div class="setting-card">
 					<h2 class="header">{{ locale.account.settings.settingCards.discord }}</h2>
 						{{#if discordUser}}
-					<p>{{ locale.account.settings.settingCards.connectedToDiscord }} {{ discordUser.username }}{{#ifCond discordUser.discriminator '#0'}}{{else}}#{{ discordUser.discriminator }}{{/ifCond}}</p>
+					<p>{{ locale.account.settings.settingCards.connectedToDiscord }} {{ discordUser.username }}
```
jonbarrow, line 148: Putting these on separate lines like this creates a space between the username and the discriminator when it is displayed

views/account/account.handlebars
```diff
@@ -145,7 +145,10 @@
...
+					{{#if discordUser.discriminator}}
```
jonbarrow, line 149: Again, see the above `#if` checks for how to actually check if a value is matching another. This always resolves to true since the value will always exist even if it's the string `0`

views/account/account.handlebars
```diff
@@ -145,7 +145,10 @@
...
+					&nbsp;#{{ discordUser.discriminator }}
```
jonbarrow, line 150: Why exactly is there an extra space here in the form of `&nbsp;`? This now creates _2_ spaces between the username and the discriminator when it is shown

jonbarrow (changes requested):
Please test your code before comitting. Somehow. Set up a local account server, run a basic handlebars server that only tests this code, etc. However it has to be done for you and your setup, it doesn't matter, but it needs to be tested before trying to commit and PR

views/account/account.handlebars
oscie57: that mustve been an accident, will remove

gitlimes: The `&nbsp;` is still there btw

jonbarrow:
Sorry about that. The branch was deleted by mistake. Reopened

PR #220, opened 2023-07-27
Docs updates: RCE patches setup, removing Tiramisu, 102-2482, remove Cemu account files

ashquarky:
I went pretty ham with the image optimisations here so if it doesn't work in Safari or smth please yell


jonbarrow:
lgtm

PR #222, opened 2023-10-04
Allow Bandwidth's messages to be dismissed

ManiacOfGitHub:
On devices with a small viewport, such as a smartphone, after Bandwidth has been clicked, you are unable to access the links in the footer without reloading the page.
pretendo network_
This PR makes it so that when you click elsewhere on the page, it will dismiss Bandwidth's prompt.


gitlimes:
LGTM! Thank you for your contribution!

PR #233, opened 2024-01-02
Network dumps docs

jonbarrow:
Adds docs for creating and submitting network dumps to help with development.

Incomplete as of right now. Missing HTTP proxy docs, ~~and Bandwidth has not been updated yet either~~.

Bandwidth has a PR open now https://github.com/PretendoNetwork/Bandwidth/pull/9

@DaniElectra Can you write the mitmproxy docs, since you use it as your proxy?
@CaramelKat Can you write the Fiddler docs, since you use it as your proxy?
@InternalLoss Can you write the Charles docs, since you use it as your proxy?

The PR is on a branch in this repo, so you can just push to this branch with your changes


docs/en_US/network-dumps.md
```diff
@@ -0,0 +1,69 @@
...
+	- [Wii U (HokakuCafe)](#wii-u-hokakucafe)
+	- [3DS (HokakuCTR)](#3ds-hokakuctr)
+	- [All (WireShark)](#all-wireshark)
+4. [HTTP Packets](#http-packets)
+	- [Fiddler](#fiddler-windows-free)
+	- [Charles](#charles-cross-platform-50)
+	- [Mitmproxy](#mitmproxy-cross-platform-free)
+
+# Security Warning
+## Network dumps will oftentimes have sensitive information, including emails, passwords, usernames, IP addresses, etc. Because of this, ensure you only provide network dumps to trusted individuals. When submitting network dumps to Pretendo Network developers your dumps are uploaded and stored in a private channel. This channel can only be accessed by a small number of people, not even all developers have access.
```
DaniElectra, line 19: We could probably use a red tip, similar to the Cemu page

imagen

docs/en_US/network-dumps.md
```diff
@@ -0,0 +1,69 @@
...
+- Wii U - To dump your NEX username and password from a Wii U connect to your Wii U with an FTP client. Navigate to `/storage_mlc/usr/save/system/act` and download all the folders inside this folder. Check the `account.dat` file in each folder and look for your NNID username in the `AccountId` field. Once found, in the same `account.dat` file locate both the `PrincipalId` and `NfsPassword` fields. Copy their values and paste them into the message of your submission.
+
+# HTTP Packets
+Some games use HTTP requests for some features. Additionally, non-game titles will often use HTTP requests for their services. The above methods will not always capture HTTP requests in a way that is usable, if at all. SpotPass data is also downloaded via HTTP requests. For this, an HTTP proxy server is required. There are several options for HTTP proxy servers, which all depend on your operating system, skill level, and, for one, your budget
+
+TODO - We have a NoSSL patch for the Wii U, but no builds are uploaded anywhere. Where should we upload this? Discord won't work anymore because of the expiring download link update. 3DS NoSSL patch by Billy https://github.com/SciresM/3DS-SSL-Patch/pull/2
```
DaniElectra, line 60: We could use a GitHub raw link to the repo: https://github.com/InternalLoss/3DS-SSL-Patch/raw/master/code.ips

A common comment should be added about requiring dumps of the console certificates

docs/en_US/network-dumps.md
```diff
@@ -0,0 +1,69 @@
...
+
+# Game Packets
+Game packet dumps are the packets sent during a session to a game server. These packets are typically encrypted, so additional data will need to be submitted in order for us to decrypt them. There are several ways to dump these packets and the data needed to decrypt them.
+
+### Wii U (HokakuCafe)
+The easiest way to dump game server packets from a Wii U is to use [HokakuCafe](https://github.com/PretendoNetwork/HokakuCafe). This is an Aroma setup module which patches IOSU to write all network frames to a `.pcap` file on the SD card.
+
+To use HokakuCafe download the [latest release](https://github.com/PretendoNetwork/HokakuCafe/releases/latest) setup module (`30_hokaku_cafe.rpx`). Place this file on the SD card at `wiiu/environments/[ENVIRONMENT]/modules/setup` and insert it back into the Wii U. Start your Wii U once, turn it off, and open the newly created `HokakuCafe/config.ini` file on the SD card. Set `Mode` to `ALL`. At the time of writing, HokakuCafe has a bug where it will not dump all packets unless `Mode` is set to `ALL`. Insert the SD card back into the Wii U and all network traffic flowing through your console should be written to a PCAP.
```
DaniElectra, line 36: > HokakuCafe has a bug

well I wouldn't say a "bug" but it's intended. By default it will only dump PRUDP packets, as that's what most Nintendo games use. Unless you mean the PRUDPv0 bug, in that case all packets should be captured with `UDP` mode too, but it doesn't hurt recommending setting the option to `ALL`.

docs/en_US/network-dumps.md
```diff
@@ -0,0 +1,69 @@
...
+
+The PCAP files are written to the `HokakuCafe` folder on the SD card, named as the date and time the session was started. HokakuCafe will also dump a BIN file, named `nexServiceToken-AAAAAAAAAA-BBBBBBBB.bin`. This file contains the data needed to decrypt the session packets, and MUST also be submitted. Otherwise the PCAP packets cannot be decrypted, and are of no use.
+
+### 3DS (HokakuCTR)
+The easiest way to dump game server packets from a 3DS is to use [HokakuCTR](https://github.com/PretendoNetwork/HokakuCTR). Like HokakuCafe this homebrew attempts to dump game traffic directly from the console. However unlike HokakuCafe, it does not dump all system traffic and may not work in all games. If HokakuCTR does not work in your game, see the last solution.
+
+To use HokakuCTR, ensure you are updated to the latest Luma release. The latest Luma release now supports 3GX plugins. Download the [latest release](https://github.com/PretendoNetwork/HokakuCTR/releases/latest). Place this file on the SD card at `luma/plugins` and rename it `default.3dx`.
```
DaniElectra, line 43:
```suggestion
To use HokakuCTR, ensure you are updated to the latest Luma release. The latest Luma release now supports 3GX plugins. Download the [latest release](https://github.com/PretendoNetwork/HokakuCTR/releases/latest). Place this file on the SD card at `luma/plugins` and rename it `default.3dsx`.
```

Universal Updater also has HokakuCTR btw: https://db.universal-team.net/3ds/hokakuctr

docs/en_US/network-dumps.md
```diff
@@ -0,0 +1,69 @@
...
+
+The PCAP is written to the SD card located at `HokakuCTR/[GAME NAME]`. HokakuCTR dumps are already decrypted, no additional data is required to decrypt them. They may be submitted as-is.
+
+### All (WireShark)
+If neither solution above works, the only option you have is to capture the traffic manually off the console. This is easiest done with WireShark and a hosted wifi access point. Create a hosted wifi access point on your PC which has WireShark installed and select that access points interface in WireShark. Connect the console to this hosted access point, and you should begin to see all the network traffic from your console in WireShark. In order to both create a hosted access point and keep your PC online, you must have 2 ways of connecting to the internet on the PC. This is easiest done by using a USB wifi adapter to host the access point, and using ethernet to remain online. Not all USB wifi adapters support hosting access points. Ensure the one you use does. [This adapter](https://www.amazon.com/dp/B07C9TYDR4) is relatively cheap and supports hosting access points.
```
DaniElectra, line 50:
```suggestion
If neither solution above works, the only option you have is to capture the traffic manually off the console. This is easiest done with WireShark and a hosted Wi-Fi access point. Create a hosted wifi access point on your PC which has WireShark installed and select that access point's interface in WireShark. Connect the console to this hosted access point, and you should begin to see all the network traffic from your console in WireShark. In order to both create a hosted access point and keep your PC online, you must have 2 ways of connecting to the internet on the PC. This is easiest done by using a USB Wi-Fi adapter to host the access point, and using ethernet to remain online. Not all USB Wi-Fi adapters support hosting access points. Ensure the one you use does. [This adapter](https://www.amazon.com/dp/B07C9TYDR4) is relatively cheap and supports hosting access points.
```

DaniElectra (commented):
Before wroking on the mitmproxy section I wanted to comment on some common parts of the docs

docs/en_US/network-dumps.md
jonbarrow: I forgot we had those, good catch

docs/en_US/network-dumps.md
jonbarrow: We don't have a public repo for the Wii U NoSSL patch. It's always just been an internal tool, it's in a private repo here https://github.com/PretendoNetwork/wiiu-nossl. Can downloads still be done from private repos if users don't have access to the repo?

docs/en_US/network-dumps.md
jonbarrow: I am talking about the v0 bug yeah, I figured just telling people to use `ALL` would be easier than explaining the actual bug. I can change it to `UDP` though, since tbh if users are using proxies we don't need anything else

MatthewL246:
In case you're interested, I created a (somewhat hacky) Dockerfile for running a mitmproxy Docker container with OpenSSL 1.1.1 [in my mitmproxy-nintendo fork](https://github.com/MatthewL246/mitmproxy-pretendo/blob/main/Dockerfile). I know that using the system-installed OpenSSL version is a common annoyance with mitmproxy. I've tested it, and it works with the Wii U and 3DS with Inkay/Nimbus enabled.

It's also published at https://ghcr.io/matthewl246/mitmproxy-pretendo. This version isn't currently suitable for network dumps because it redirects requests from Nintendo to Pretendo by default, but this could be changed easily if you're interested.

jonbarrow:
Using the system OpenSSL is indeed the biggest annoyance I've had with using mitmproxy for these console and is the biggest reason I stick with Charles. I'm definitely interested in something like this. Would you be willing to also PR these changes back into the upstream repo?

@SuperMarioDaBom and I have been working towards Dockerizing all of our services anyway

jonbarrow:
Bandwidth officially has a PR open to add his side of things https://github.com/PretendoNetwork/Bandwidth/pull/9

jonbarrow:
@hauntii I believe you also use Fiddler, correct? I know Jemma has a full time job outside of this. Would you mind contributing to the Fiddler docs in her absence?

jonbarrow:
@hauntii I believe you also use Fiddler, correct? I know Jemma has a full time job outside of this. Would you mind contributing to the Fiddler docs in her absence?

Or maybe we should scrap all of that and only focus on mitmproxy, seeing as it's free and cross platform, and @MatthewL246 has been doing some upkeep on our old wrapper for it including that Dockerfile. Using that should make things pretty easy for new users, since it only really requires using Docker?

MatthewL246:
Sure, I would be interested in making a PR for that. I was thinking that I could set up builds for multiple versions of the container using different tags - one that redirects Nintendo to Pretendo as originally intended, one for Wii U packet captures (doesn't modify any traffic and sends the Wii U client certificate), and one for 3DS packet captures (sends the CTR client certificate).

docs/en_US/network-dumps.md
DaniElectra: oh my bad, I didn't notice you are talking about the Wii U patch. You can't download from private repos, but what are your thoughts on opening that repo to the public? I don't think it needs to be hidden if we are going to distribute the patch anyway

jonbarrow:
> Sure, I would be interested in making a PR for that. I was thinking that I could set up builds for multiple versions of the container using different tags - one that redirects Nintendo to Pretendo as originally intended, one for Wii U packet captures (doesn't modify any traffic and sends the Wii U client certificate), and one for 3DS packet captures (sends the CTR client certificate).

This works for me. Though for less tech-savvy users, would it be possible to still provide a "combined" version that doesn't modify traffic? So that way users can capture traffic from both consoles at once?

I've had people who needed help even just *installing* software like Fiddler, which has an installation wizard. So anything we can do to help minimize friction is a plus in my book. Having things be separated would be useful for development though on our end

> Update: Here's an example network dump generated by `mitmdump -w flows.txt`: [flows.txt](https://github.com/PretendoNetwork/website/files/13820240/flows.txt). It can easily be loaded by running `mitmweb -n` to start the web UI without starting the proxy server and opening the file.

Awesome! I gave your container a try just now and it seems to run well. The only thing I'm concerned about is exporting the traffic. I see you uploaded your flows as `flows.txt`, but the web UI (which is what I'll be having users use in these docs) exports as an extensionless `flows` file. This works fine for us, but wouldn't be able to be submitted through our Discord bot. Our Discord bot checks for network dumps by their extension, which would break here (unless we also check for a file just named `flows`? But then anyone who renames the file, which is something we suggest to do, wouldn't be able to submit them)

Does mitmproxy support exporting as a HAR in the web UI? I remember hearing that this was coming at some point a while ago but I'm unsure if this ever happened. Our Discord bot is already setup to detect HAR dumps so that would fix the issue

DaniElectra:
You can export mitmproxy dumps to HAR by specifying the output file on the `hardump` option. More info: https://mitmproxy.org/posts/har-support/

jonbarrow:
> on the `hardump` option

I did see this, but this only seems to apply to when using the CLI. I don't see anywhere that mentions support for exporting in the web UI, which is what these docs will focus on as a GUI is likely easier for random people rather than dealing with a CLI

MatthewL246:
Hardump can be enabled in the config file. It would apply to all 3 of the mitmproxy modes.

jonbarrow:
That sounds perfect then :+1:

DaniElectra:
The option is on the GUI too 

imagen

jonbarrow:
You're right, I missed that. I didn't see it mentioned in the changelog so I didn't think to look for it there. We can either enable it in the config automatically or have the docs tell users to set this. Either works for me, whatever you think would be easiest

MatthewL246:
The HAR doesn't seem to include all of the information that the flows file includes (like the certificate info), but I guess it has what you need? I can absolutely enable `hardump` by default in the network dumping config.

jonbarrow:
So long as it has the headers, paths, bodies, etc, that should be enough for our needs? @DaniElectra what say you

MatthewL246:
> Though for less tech-savvy users, would it be possible to still provide a "combined" version that doesn't modify traffic? So that way users can capture traffic from both consoles at once?

I'm not sure. Mitmproxy does provide a way to specify a different client certificate per domain, but this would require a comprehensive list of every Nintendo domain that requires a client certificate. It also would have the limitation that only one client certificate could be specified per domain. Is there any domain that both the Wii U and 3DS need to connect to and requires a different client certificate depending on the console? That would be a problem.

jonbarrow:
> I think it would be easier to just tell people to run `docker run mitmproxy-pretendo:wiiu` or `docker run mitmproxy-pretendo:3ds` depending on the console.

That sounds reasonable to me given the circumstances. I'm not that familiar with mitmproxy so I wasn't sure how possible something like that was

DaniElectra:
> So long as it has the headers, paths, bodies, etc, that should be enough for our needs? @DaniElectra what say you

All the relevant data is present, so I think it should be enough

DaniElectra:
> Mitmproxy does provide a way to specify a different client certificate per domain, but this would require a comprehensive list of every Nintendo domain that requires a client certificate

We probably need this anyway for the Wii U as it has a different "common" and "account" client certificate

docs/en_US/network-dumps.md
jonbarrow: I asked Quarky earlier today about why it was private. Waiting on a reply back before opening it back up. For now both SSL patches are included in the public mitmproxy repo

jonbarrow: The answer is because it's a development tool and may confuse regular users who just install everything from the org

jonbarrow:
@DaniElectra I added the final bit of docs. I believe this should be everything? I did notice however in my Wii U testing that POST requests to the SOAP server did not save their request bodies. I am unsure if that is a bug in mitmproxy or a fluke on my end but it happened with all SOAP requests in my test capture. The `Content-Length` header is present and claims to have sent data, but none is saved. Since we don't really support the SOAP server anyway, and it's required for the eShop so I doubt it will go down in April, I think this is fine? Just something we need to keep an eye on in the future I suppose

DaniElectra:
>  did notice however in my Wii U testing that POST requests to the SOAP server did not save their request bodies

That's weird, I tested on the 3DS and the SOAP request bodies are being saved. Can you send me the dump on discord?

docs/en_US/network-dumps.md
```diff
@@ -0,0 +1,137 @@
...
+Install Docker for your operating system using the official [setup guide](https://docs.docker.com/get-docker/). Then follow the steps for your console type.
+
+## 3DS
+1. Download [this IPS patch](https://github.com/PretendoNetwork/mitmproxy-nintendo/raw/master/ssl-patches/0004013000002F02.ips). This IPS patch patches the SSL sysmodule to disable SSL verification, allowing the console to connect to the proxy server with TLS connections.
+2. Place the patch on your SD card at `SD:/luma/sysmodules/0004013000002F02.ips`.
+3. Place the SD card back into your 3DS.
+4. Launch into the Luma settings by holding `SELECT` while powering on.
+5. Ensure both `Enable loading external FIRMS and modules` and `Enable game patching` are enabled before booting.
+6. Launch Nimbus and connect to Nintendo Network.
+7. On your computer, create a `3ds-dumps` directory and run the following command in a command prompt: `docker run -it --rm -p 8083:8083 -v ./3ds-dumps:/home/mitmproxy/dumps ghcr.io/pretendonetwork/mitmproxy-nintendo:3ds mitmdump`
```
DaniElectra, line 71: We should probably specify that the folder should be created on the same directory as the terminal, or make the user create the folder from the terminal with `mkdir 3ds-dumps`

jonbarrow:
I sent it in the dumps channel

DaniElectra (commented):
LGTM, only one thing

docs/en_US/network-dumps.md
jonbarrow: Does `mkdir` work on all platforms?

docs/en_US/network-dumps.md
```diff
@@ -0,0 +1,137 @@
...
+13. Select `Ok` then `Save` and run the connection test. Your 3DS should connect to the internet and you should see connections being made in the proxy server
+14. See the end of this section for final steps.
+
+## Wii U
+1. Download [this Aroma setup module](https://github.com/PretendoNetwork/mitmproxy-nintendo/raw/master/ssl-patches/30_nossl.rpx). This patches the SSL sysmodule to disable SSL verification, allowing the console to connect to the proxy server with TLS connections.
+2. Place the patch on your SD card at `SD:/wiiu/environments/aroma/modules/setup/30_nossl.rpx`. If there are other patches with the same ID `30`, this is fine.
+3. Place the SD card back into your Wii U and turn on the console.
+4. Launch into the Aroma settings by pressing `L + DPAD-DOWN + SELECT`.
+5. Enter `Inkay > Patching` and toggle `Connect to the Pretendo Network` to ***FALSE***.
+6. On your computer, create a `wiiu-dumps` directory and run the following command in a command prompt: `docker run -it --rm -p 8082:8082 -v ./wiiu-dumps:/home/mitmproxy/dumps ghcr.io/pretendonetwork/mitmproxy-nintendo:wiiu mitmdump`
```
MatthewL246, line 86: When using the Windows command line, the volume needs to be `-v .\wiiu-dumps:/home/mitmproxy/dumps` (with the backslash for the source path). Alternatively, you could tell Windows users to run `wsl docker run -it --rm -p 8082:8082 -v ./wiiu-dumps:/home/mitmproxy/dumps ghcr.io/pretendonetwork/mitmproxy-nintendo:wiiu mitmdump` to run the command inside WSL, where the forward slash bind mount does work.

docs/en_US/network-dumps.md
DaniElectra: yep, it's standard through unix platforms and it's also available on windows (even on the classic CMD)

jonbarrow: Perfect :+1:. I was mostly worried about Windows

docs/en_US/network-dumps.md
jonbarrow: Is WSL installed and enabled by default on Windows? These docs are aiming to cater to brand new users who may have never touched even a command prompt before. I've had to hold peoples hands while they installed software like Fiddler, which has an installation wizard, so I'd like to avoid overwhelming these users

MatthewL246: No, it's not enabled by default. The thing is, Docker Desktop for Windows [requires WSL 2 to be set up](https://docs.docker.com/desktop/install/windows-install/) because the Docker daemon itself runs inside WSL. So, anyone with Docker installed on Windows will necessarily have WSL2 enabled.

DaniElectra: > Docker Desktop for Windows [requires WSL 2 to be set up](https://docs.docker.com/desktop/install/windows-install/)

not necessarily, there is another option for using the Hyper-V backend, which presumably doesn't use WSL? haven't tested though

MatthewL246: I think WSL was the default option when I installed it. Unfortunately installing WSL2 and Docker Desktop for Windows might require some hand-holding.

jonbarrow: > > Docker Desktop for Windows [requires WSL 2 to be set up](https://docs.docker.com/desktop/install/windows-install/)
> 
> not necessarily, there is another option for using the Hyper-V backend, which presumably doesn't use WSL? haven't tested though

The Hyper-V backend is for running Windows containers I thought? Since this runs a Linux container users would likely still need the WSL backend. I'll update the docs to clarify that users need the WSL version

jonbarrow:
@MatthewL246 @DaniElectra Aight, added WSL notes and updated the commands. Should be good?

MatthewL246:
I think it should be good.

DaniElectra (approved):
Looks good

MatthewL246 (approved):

PR #250, opened 2024-02-20
Added Filter Progress

MythicalPlayz:
This PR (which I will trun into a draft) adds the filter Progress like in the Figma Design
The Filters then can also be localised if needed
Old:
Screenshot 2024-02-17 214435
New:
Screenshot 2024-02-17 214507

### DIFF
--- a/public/assets/css/progress.css
+++ b/public/assets/css/progress.css
@@ -13,20 +13,38 @@ MOVE PROGRESS CSS HERE
 }
 
 .progress-filter {
-	padding: 8px 72px;
+	padding: 8px;
     border-radius: 10px;
     background: var(--bg-shade-0);
     grid-column: span 2;
 }
 
 .progress-filter .container {
-	display: grid;
-	grid-template-columns: repeat(6, 1fr);
-	grid-gap: 20px;
+	display: flex;
+	justify-content: flex-start;
 }
 
-.progress-filter .container button {
-	padding: 0px;
+.progress-filter .container .sub-container {
+	display: flex;
+	margin-inline: 5px;
+	position: relative;
+}
+
+.progress-filter .container .sub-container:not(:last-child):after {
+	content: '';
+	position: absolute;
+	top: 10%;
+	bottom: 10%;
+	right: -10px;
+	width: 3px;
+	background-color: var(--bg-shade-3);
+}
+
+
+
+.progress-filter .container .sub-container button {
+	padding: 0px 8px;
+	margin-inline-start: 5px;
 }
 
 .off {
@@ -39,9 +57,20 @@ MOVE PROGRESS CSS HERE
 		grid-column: span 1;
 		padding: 10px;
 	}
+}
+@media screen and (max-width: 576px) {
 
 	.progress-filter .container {
-		grid-template-columns: repeat(3, 1fr);
-		grid-template-rows: repeat(2, 1fr);
+		justify-content: center;
 	}
+
+	.progress-filter .container .sub-container {
+		margin: 0px;
+	}
+
+	.progress-filter .container .sub-container:not(:last-child):after {
+		content: '';
+		display: none;
+	}
+
 }
--- a/views/progress.handlebars
+++ b/views/progress.handlebars
@@ -26,12 +26,18 @@
 
 		<div class="progress-filter">
 			<div class="container" id="filter-container">
-				<button class="filter primary" id="none"> {{locale.filter.none}} </button>
-				<button class="filter" id="wii u"> {{locale.filter.wiiu}} </button>
-				<button class="filter" id="3ds"> {{locale.filter.3ds}} </button>
-				<button class="filter" id="web"> {{locale.filter.web}} </button>
-				<button class="filter" id="game"> {{locale.filter.game}} </button>
-				<button class="filter" id="service"> {{locale.filter.service}} </button>
+				<div class="sub-container" id="no-filter">
+					<button class="filter primary breaker" id="none"> {{locale.filter.none}} </button>
+				</div>
+				<div class="sub-container" id="device">
+					<button class="filter" id="wii u"> {{locale.filter.wiiu}} </button>
+					<button class="filter" id="3ds"> {{locale.filter.3ds}} </button>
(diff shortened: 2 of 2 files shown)


public/assets/js/filter-progress.js
```diff
@@ -0,0 +1,24 @@
...
+const buttons = document.getElementById('filter-container').children
+let selectedButton = buttons[0]
+for (const button of buttons){
+    button.addEventListener('click',(event) => {
+        selectedButton.classList.remove('primary')
+        selectedButton = event.target
+        selectedButton.classList.add('primary')
+        showWithFilter(event.target.id);
+    })
+}
```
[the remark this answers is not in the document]
jonbarrow: I can't make multiple suggestions over the same lines in one review, but I would also prefer to not use the `children` property on `filter-container`. Instead using `querySelectorAll` targeting the buttons would be safer in the event that we need to restructure the HTML and not worry about updating the JS

src/cache.js
```diff
@@ -69,6 +69,25 @@ query getProjectsV2Fields($id: ID!, $cursor: String!) {
...
+            }
+        }
+        master: object(expression: "master:README.md") {
+            ... on Blob {
+                text
+            }
+        }        
+    }
+}
+`; //Loophole to get the ReadMe No Matter the branch name
```
jonbarrow, line 88: The website is behind on this, but we've been moving to the comment format made by [better comments](https://marketplace.visualstudio.com/items?itemName=aaron-bond.better-comments)

```suggestion
`; // * Loophole to get the ReadMe No Matter the branch name
```

src/cache.js
```diff
@@ -69,6 +69,25 @@ query getProjectsV2Fields($id: ID!, $cursor: String!) {
...
+        main: object(expression: "main:README.md") {
+            ... on Blob {
+                text
+            }
+        }
+        master: object(expression: "master:README.md") {
+            ... on Blob {
+                text
+            }
+        }
```
jonbarrow, line 85: You don't have to do this in 2 steps. Target `HEAD` rather than a branch and it gives you the repositories default branch

```suggestion
       readme: object(expression: "HEAD:README.md") {
            ... on Blob {
                text
            }
        }        
```

src/cache.js
```diff
@@ -69,6 +69,25 @@ query getProjectsV2Fields($id: ID!, $cursor: String!) {
...
+        master: object(expression: "master:README.md") {
+            ... on Blob {
+                text
+            }
+        }        
+    }
+}
+`; //Loophole to get the ReadMe No Matter the branch name
+
+const serviceApps = ["account","nex","friends","wii u chat", "juxt", "website"]
```
jonbarrow, line 90:
```suggestion
const serviceApps = ["account", "nex", "friends", "wii u chat",  "juxt",  "website"];
```

This could also probably go on multiple lines, but I won't require it

src/cache.js
```diff
@@ -133,6 +152,69 @@ async function getGitHubProjectsV2Fields(id, after='') {
...
+	}
+	else if (description.includes('3ds') || readMe.includes('3ds') || title.includes('3ds')){
+		types.push('3DS')
+	}
+	else{
+		types.push('Wii U')
+	}
+	
+	return types
+}
```
jonbarrow, line 216:
```suggestion
async function getRepoType(name, title) {
	const data = await github.request(getRepositoryDescription, {
		orgName: 'PretendoNetwork',
		repoName: name
	});

	const readme = data.repository.readme.text.split('\n')[0].toLowerCase();
	const description = data.repository.description?.toLowerCase() || '';

	return setRepoType(title.toLowerCase(), description, readme);
}


function setRepoType(title, description, readMe) {
	const types = [];
	let isGame = true;

	for (const app of serviceApps) {
		if (title.includes(app)) {
			types.push('Service');
			isGame = false;
			break;
		}
	}

	if (isGame) {
		types.push('Game');
	}

	if (title.includes('(') && isGame) {
		if (title.includes('3ds')) {
			types.push('3DS');
		} else {
			types.push('Wii U');
		}

		return types;
	}

	if (title === 'nex' || title.includes('juxt') || title.includes('account')) {
		types.push('3DS');
		types.push('Wii U');

		if (title.includes('juxt')) {
			types.push('Website');
		}
	} else if (title.includes('web')) {
		types.push('Website');
	} else if (description.includes('3ds') || readMe.includes('3ds') || title.includes('3ds')) {
		types.push('3DS');
	} else{
		types.push('Wii U');
	}

	return types;
}
```

src/cache.js
```diff
@@ -69,6 +69,25 @@ query getProjectsV2Fields($id: ID!, $cursor: String!) {
 	}
 }
 `;
+const GetRepositoryDescription = gql`
+query GetRepositoryDescription($orgName: String!, $repoName: String!){
```
jonbarrow, line 73:
```suggestion
const getRepositoryDescription = gql`
query getRepositoryDescription($orgName: String!, $repoName: String!){
```

src/cache.js
```diff
@@ -172,7 +254,8 @@ async function updateGithubProjectsCache() {
 				done: [],
 				in_progress: [],
 				todo: []
-			}
+			},
+			type: []
```
jonbarrow, line 258:
```suggestion
			types: []
```

src/cache.js
```diff
@@ -181,6 +264,9 @@ async function updateGithubProjectsCache() {
 			extractedData.cards[field.column.toLowerCase().replace(' ', '_')]?.push(field.title);
 		}
 
+		const types = await getRepoType(project.url.split("/")[4], project.title)
+		extractedData.type = types
```
jonbarrow, line 268:
```suggestion
		extractedData.types = await getRepoType(project.url.split("/")[4], project.title);
```

Or just set this directly on the object rather than reassigning it here

jonbarrow (changes requested):
Mostly just style and consistency changes. Most of this will be redone/removed in the future so I'm not too worried about the finer details right now. Once the admin panel is done being rewritten I planned to add a kanban board to it and ditch GitHub Projects entirely, so once that's done we can do all this tagging and filtering and stuff in  a more streamlined way

PR #257, opened 2024-03-22
Add wii u console requirement & Online guide setup for the cemu page

VasilisThePikachu:
I had to figure this out the hard way with a friend since the documentation for Cemu is surprisingly lacking. Saw this repo was open and thought I would try to clarify the docs.

Also for anyone else wanting to do this with a real wii u. I have also included a nice link to the Cemu guide for setting up online play.

### DIFF
--- a/docs/en_US/install/cemu.md
+++ b/docs/en_US/install/cemu.md
@@ -2,21 +2,15 @@
 
 # Cemu
 
-<div class="tip red">
-	<strong>Note:</strong>
-	Currently, the usage of a real modded Wii U Console is required to dump the online files from your Pretendo user.
-</div>
-
 ## Download
 <div class="tip red">
 	<strong>Note:</strong>
 	Only experimental builds of Cemu 2.0 are supported. At this time Cemu does not have a stable release of Cemu 2.0 which supports Pretendo
 </div>
 
-
 Cemu 2.0 has official built-in support for Pretendo as of October 10, 2022. Head to the Cemu GitHub [releases](https://github.com/cemu-project/Cemu/releases) page and download the latest Cemu experimental release (tagged as `Pre-release`). Only `Cemu 2.0-5 (Experimental)` and above is supported at the moment. Additionally you may build Cemu from source using the provided [build instructions](https://github.com/cemu-project/Cemu/blob/main/BUILD.md)
 
-After you can follow [Cemu's online guide](https://cemu.cfw.guide/online-play.html) to dump the online files from your Pretendo user account on your Wii U.
+After you can follow [Cemu's online guide](https://cemu.cfw.guide/online-play.html) to dump the online files from your Wii U after you logged in to your Pretendo Account on it.
 
 ## Connecting to Pretendo
 Once you have Cemu setup for online play navigate to `Options > General settings > Account`. You should now see a section titled `Network Service`. Select your PNID from the `Active account` menu and select the `Pretendo` Network Service option. Cemu should now be connected to Pretendo's servers
--- a/docs/en_US/install/cemu.md
+++ b/docs/en_US/install/cemu.md
@@ -2,13 +2,21 @@
 
 # Cemu
 
+<div class="tip red">
+	<strong>Note:</strong>
+	Currently, the usage of a real modded Wii U Console is required to dump the online files from your Pretendo user.
+</div>
+
 ## Download
 <div class="tip red">
 	<strong>Note:</strong>
 	Only experimental builds of Cemu 2.0 are supported. At this time Cemu does not have a stable release of Cemu 2.0 which supports Pretendo
 </div>
 
+
 Cemu 2.0 has official built-in support for Pretendo as of October 10, 2022. Head to the Cemu GitHub [releases](https://github.com/cemu-project/Cemu/releases) page and download the latest Cemu experimental release (tagged as `Pre-release`). Only `Cemu 2.0-5 (Experimental)` and above is supported at the moment. Additionally you may build Cemu from source using the provided [build instructions](https://github.com/cemu-project/Cemu/blob/main/BUILD.md)
+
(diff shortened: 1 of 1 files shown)


jonbarrow:
The link the Cemu's guide is fine, but the note about needing a real Wii U is both unnecessary and somewhat incorrect.

The guide tells you to follow Cemu's guide for online play. Cemu's guide lays out those requirements, and makes it clear you need a Wii U. Adding it to our guide as well is unnecessary in this case, as Cemu's guide already says this.

Additionally the notice you proposed says:

> Currently

Saying "currently" is incorrect, as we do not intend to allow for connecting to our servers without the use of a real console, including when using emulators. This is done for security purposes, and we do not foresee this changing at any time.

> the documentation for Cemu is surprisingly lacking

This is, honestly, a bit intentional. There's nothing really to document on our end besides just:

- The Cemu version you need
- Saying you still need to follow Cemu's guide
- Pointing out where the Pretendo setting is (because that's not in Cemu's guide)
- Mentioning the state of Miiverse in Cemu

All of which the guide does currently do. The bulk of the information is on Cemu's side, not ours, which is why we tell users to go do Cemu's guide *first*. There's nothing really else for us to say on the subject, and trying to re-post information from Cemu's guide would have the potential to make our guide outdated and/or conflict with the official guide, neither of which we want

VasilisThePikachu:
> Additionally the notice you proposed says:
> 
> > Currently
> 
> Saying "currently" is incorrect, as we do not intend to allow for connecting to our servers without the use of a real console, including when using emulators. This is done for security purposes, and we do not foresee this changing at any time.

I was not aware that this is the case, I was under the assumption that at one point it would be allowed to preserve online play. But I agree its probably better to remove the "currently"

> > the documentation for Cemu is surprisingly lacking
> 
> This is, honestly, a bit intentional. There's nothing really to document on our end besides just:
>
> All of which the guide does currently do. The bulk of the information is on Cemu's side, not ours, which is why we tell users to go do Cemu's guide _first_. There's nothing really else for us to say on the subject, and trying to re-post information from Cemu's guide would have the potential to make our guide outdated and/or conflict with the official guide, neither of which we want. Our guide is intentionally left kinda bare when it comes to Cemu

Totally understandable. I will make some changes soon after posting this.

jonbarrow:
> I was under the assumption that at one point it would be allowed to preserve online play

It is possible and we allowed this in the past, but we removed this out of security concerns. In order to implement proper system bans, we ban the console certificates linked to each console. These certificates are unique, and signed by Nintendo. No one can create new ones without Nintendo's private key. Without a real console to dump from, you won't have one of these certificates. Which means there's no real way to ban emulator-only users.

Account bans are worked around with alts. IP bans are worked around with VPNs and rolling IPs. Provisioning our own certificates is not feasible as we have to decide how/when to provision them. One per account? Alts makes this pointless. One per IP? VPNs and rolling IPs make this useless. Someone would always have a way to get around the ban, whether that be through making an alt account, changing their IP, or getting more of our own provisioned certificates. There's just no good long-term solution there.

The only truly effective way to do this is by banning the real, legitimate, certificate, since we can validate the certificate is legitimate (we have the public key), and no one can generate illegitimate ones (no one but Nintendo has the private key).

Additionally, with the Yuzu/Citra situation happening we are *definitely* not interested in allowing this, as it could very likely be argued that we would be directly supporting the use of pirated games (since if they don't have a console, they very likely didn't buy the games either). We know it is unlikely Nintendo would act on us because of this, but we take these precautions to not step on any toes.

DaniElectra:
The target branch needs to be changed from `master` to `dev`

jonbarrow:
> The target branch needs to be changed from `master` to `dev`

Thank you, I totally missed that

docs/en_US/install/cemu.md
```diff
@@ -10,10 +10,12 @@
 
 Cemu 2.0 has official built-in support for Pretendo as of October 10, 2022. Head to the Cemu GitHub [releases](https://github.com/cemu-project/Cemu/releases) page and download the latest Cemu experimental release (tagged as `Pre-release`). Only `Cemu 2.0-5 (Experimental)` and above is supported at the moment. Additionally you may build Cemu from source using the provided [build instructions](https://github.com/cemu-project/Cemu/blob/main/BUILD.md)
 
+After you can follow [Cemu's online guide](https://cemu.cfw.guide/online-play.html) to dump the online files from your Wii U after you logged in to your Pretendo Account on it.
```
jonbarrow, line 13:
```suggestion
Ensure you have followed [Cemu's guide](https://cemu.cfw.guide/online-play.html) to set up the emulator for online play. When dumping your user account files, ensure you select your PNID.
```

docs/en_US/install/cemu.md
```diff
@@ -10,10 +10,12 @@
...
+
 ## Connecting to Pretendo
 Once you have Cemu setup for online play navigate to `Options > General settings > Account`. You should now see a section titled `Network Service`. Select your PNID from the `Active account` menu and select the `Pretendo` Network Service option. Cemu should now be connected to Pretendo's servers
 
 <center><img src="/assets/images/docs/install/cemu/network-services-settings.png"/></center>
 
 ## Miiverse
-Cemu has limited to no Miiverse support as of now. Currently no Miiverse features, including the Miiverse applet as well as Miiverse features in games, will function.
\ No newline at end of file
+Cemu has limited to no Miiverse support as of now. Currently no Miiverse features, including the Miiverse applet as well as Miiverse features in games, will function.
```
jonbarrow, line 21: This is no longer true, some Miiverse stuff does work and you *can* get Cemu to load the applet if you modify it

```suggestion
Cemu has limited to no Miiverse support as of now. Some in game features may work, but this is not guaranteed. The Miiverse applet does not work in official builds.
```

jonbarrow (changes requested):
Information is good, just some rewording to make it sound more like the rest of the page

PR #260, opened 2024-03-27
Chore: update deps

Jvr2022:


jonbarrow:
Sorry about that. The branch was deleted by mistake. Reopened

jonbarrow:
As is, this PR would break the website. There are several major version updates which introduce breaking changes (such as the jump from `got` 11 to 14). I'm going to close this for now, as there's not a HUGE reason to update these atm and they can't be merged without other, larger, changes to the codebase

Please test changes before making PRs in the future

PR #264, opened 2024-04-08
Fix old discord images

Auxy6858:
Fixed an issue where images obtained from discord would not load


jonbarrow:
Good catch. Though it would be better if the images were saved locally, like they are for other pages

Auxy6858:
If you want I'll download the images and add them to the public/assets/images folder and reference them

PR #273, opened 2024-04-12
Japanese translation fixed.

reindex-ot:
Japanese translation was wrong, so I fixed it.


jonbarrow:
You're on the wrong branch. Target dev not master

PR #278, opened 2024-04-12
Update 3ds.md

Yevnilc-S-C:
Add information on account linking and creation, and add warning for an error that occurs that says the email's domain name is invalid, and added information that tells users that it's safe to ignore.

I tried to format it as closely as I could, tell me if there's anything that might need to be fixed, thanks!
-Sarah


jonbarrow:
The information added here is unnecessary. The previous line tells the user to go through the account setup process as if it were a NNID. Telling the user to go through the settings app and such is just restating what was already said, that was clear already

As for the email issue, it is a temporary issue and not something which I feel needs to be added to the docs. Trying to constantly maintain, adding and removing, all temporary errors is not something I feel we should spend our time on

In the future you may want to open an issue before making changes and a PR so that whether or not those changes are necessary can be discussed

Yevnilc-S-C:
I'm very sorry for that, I'll make sure to open up an issue for next time

PR #282, opened 2024-04-17
[WIP] Game support page

ashquarky:
A new page showing the supported games and whether they're online, tester-only or offline. I'm half-asleep writing this so I know the code isn't great, but we really need to put this information up *somewhere* for people to reference. Game info is just hard-coded into the server for now, we can make a big fancy API like the Progress page some other day. I don't expect we'd list every unsupported game, just the ones people ask about.

This PR is incomplete:
- Not every game is added
- It's not hooked into the navigation bars or anything

Posting early in case a fundamental overhaul is needed and to make sure this direction is sensible.

### DIFF
--- a/locales/en_GB.json
+++ b/locales/en_GB.json
@@ -139,6 +139,10 @@
 		"title": "Our progress",
 		"description": "Check the project's progress and goals! (Updated every hour or so, does not reflect ALL project goals or progress)"
 	},
+	"gamesPage": {
+		"title": "Game support",
+		"description": "Check which games are supported and online."
+	},
 	"blogPage": {
 		"title": "Blog",
 		"description": "The latest updates in condensed chunks. If you want to see more frequent updates, consider <a href=\"/account/upgrade\" target=\"_blank\">supporting us</a>.",
--- a/locales/en_US.json
+++ b/locales/en_US.json
@@ -252,6 +252,10 @@
 		"title": "Our progress",
 		"description": "Check the project progress and goals! (Updated every hour or so, does not reflect ALL project goals or progress)"
 	},
+	"gamesPage": {
+		"title": "Game support",
+		"description": "Check which games are supported and online."
+	},
 	"blogPage": {
 		"title": "Blog",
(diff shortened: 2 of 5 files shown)


jonbarrow:
> A new page showing the supported games

This would be the progress page, it's what it's designed for. To show the list of supported games and what features are/are not implemented in them. I think having status pages and such is a good idea but why duplicate this information across multiple pages?

ashquarky:
The progress page shows the state of our code, but not the live servers - what's public, what's tester-only, what's unavailable, and eventually with the addition of an API what's offline for maintenance or what's crashed. More of a server status page than anything?

We could definitely integrate that information into progress though, maybe with little badges over the game or something.

jonbarrow:
> The progress page shows the state of our code, but not the live servers

The state of our code *is* what is live in our servers. As things get implemented, they get deployed, so it's the same thing. In fact that's the whole purpose of the progress page. It wasn't made with "this is what our code is up to" in mind, it was made with "here's a general list of features you can and cannot do in games" in mind

> what's public, what's tester-only, what's unavailable

If a feature is unavailable it's marked as such already. If a game is unavailable it just doesn't have a card. That already seems pretty straight forward?

As for the public vs testing stuff, I've mentioned this in other issues/PRs/on Discord but the reason why there's no list of public vs beta games is because the status of a server can, and does, change at any time. So trying to maintain a list like that would be annoying at best

> what's offline for maintenance or what's crashed. More of a server status page than anything?

This sounds like a slightly different issue than originally proposed. "Supported games" is already covered by the progress page, but "server status" is not and I do agree it should probably have it's own page. Kinda like how we have https://stats.uptimerobot.com/R7E4wiGjJq/ though this seems outdated and busted

> We could definitely integrate that information into progress though, maybe with little badges over the game or something

If there's overlap in the information/scope then we should probably think about consolidating it yeah. Like I said the progress page is due for an overhaul to begin with so now might be a good time to do that? Kill 2 birds with one stone here? @hauntii @gitlimes what do y'all think?

MisterSheeple:
Any updates on this? As a longtime user, I've always wondered why there hasn't been a page showing a full list of what games are supported. It's kind of weird that an unofficial wiki is the main source for that.

PR #283, opened 2024-04-17
Update FAQ entries for public release and SSSL

ashquarky:
Some of the FAQ entries are pretty far out of date and give people the wrong information. Update these to mention that we have public, free servers, and talk about Aroma and SSSL instead of Coldboot Haxchi.

### DIFF
--- a/locales/en_GB.json
+++ b/locales/en_GB.json
@@ -55,7 +55,7 @@
 			},
 			{
 				"question": "How do I use Pretendo?",
-				"answer": "Pretendo is currently not in a state that is ready for public use. However, once it is, you will be able to use Pretendo simply by running our homebrew patcher on your console."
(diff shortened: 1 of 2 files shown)


ashquarky:
I have no idea what needs to happen for Weblate to be aware of these changes, btw

jonbarrow:
LGTM. @hauntii @gitlimes can you confirm that Weblate won't implode with these changes?

jonbarrow (approved):

gitlimes:
It shouldn't, no!

jonbarrow:
@gitlimes I forgot to ask this the other day, but how does this affect other languages? Will this invalidate the other translations? Will they remain outdated? How does that work with Weblate

gitlimes:
They will be shown as outdated on the website interface. There is also an option in the we late interface to delete outdated translation strings from the locale files, which should probably be used.

PR #284, opened 2024-04-19
docs/3ds: Don't use "At a later date" - confusing wording

ashquarky:
Some users were reading "you can link your PNID at a later date" to mean that PNID linking is *unimplemented*. Update the wording to reflect the truth - you can link it whenever you want!


jonbarrow:
lgtm

PR #286, opened 2024-04-21
Add password length requirements on registration page

angeloanan:
Had some trouble creating a Pretendo Account due to password length hints not being sent to the client.

This PR adds the said minimum / maximum requirements of registration to the page.
The number of characters were derived from: https://github.com/PretendoNetwork/account/blob/8c07ae82f90e26424c39b53c9ea73ef7cdaee731/src/services/api/routes/v1/register.js#L164-L178

### DIFF
--- a/views/account/register.handlebars
+++ b/views/account/register.handlebars
@@ -25,11 +25,11 @@
 			</div>
 			<div>
 				<label for="password">{{ locale.account.loginForm.password }}</label>
-				<input name="password" id="password" type="password" autocomplete="new-password" required>
+				<input name="password" id="password" type="password" autocomplete="new-password" minlength="6" maxlength="16" required>
 			</div>
 			<div>
 				<label for="password_confirm">{{ locale.account.loginForm.confirmPassword }}</label>
-				<input name="password_confirm" id="password_confirm" type="password" autocomplete="new-password" required>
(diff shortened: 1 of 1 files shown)


angeloanan:
As a side note / discussion: Why do Pretendo have a maximum password length requirements?
It seems to me that its a backwards behavior as password managers are becoming more popular.

vonmillhausen:
> As a side note / discussion: Why do Pretendo have a maximum password length requirements? It seems to me that its a backwards behavior as password managers are becoming more popular.

On the 3DS, the device's UI limits you to entering a maximum of 16 characters for the password field. Not sure about the Wii U.

jonbarrow:
> Why do Pretendo have a maximum password length requirements?

- https://en-americas-support.nintendo.com/app/answers/detail/a_id/1087/~/how-to-create-a-nintendo-network-id-on-wii-u
- https://en-americas-support.nintendo.com/app/answers/detail/a_id/2221/~/how-to-create-a-nintendo-network-id-on-nintendo-3ds

"The password you choose should be between 6 and 16 characters long."

This is why. This is the limit enforced by Nintendo, both on their original servers and on the devices themselves. There is a physical limit to how many characters you can enter on the consoles, allowing people to create passwords larger than that on the website would render them useless

PR #287, opened 2024-04-21
docs(install/3ds): rewrite `Signing into your PNID` section

angeloanan:
This PR improves the Pretendo Network 3DS installation guide by rewriting the `Signing into your PNID` section.

More specifically:
* It actually includes the step on how to create or link a Pretendo account **IN** the console
* It highlights a known Nimbus issue where initial installation and profile switch might show an already-linked Nintendo ID profile (https://github.com/PretendoNetwork/nimbus/issues/32)
* It rewrote and reorders paragraphs to be less wordy.

CleanShot 2024-04-21 at 11 23 00

### DIFF
--- a/docs/en_US/install/3ds.md
+++ b/docs/en_US/install/3ds.md
@@ -83,18 +83,24 @@ Depending on how you installed Nimbus, launch it either through the Homebrew Lau
 
 Your selection persists between reboots.
 
-## Signing into your PNID
+## Signing in into Pretendo Network
 
-The 3DS does not rely on NNIDs for the vast majority of it's game servers. Because of this, using a PNID is also not required for most games<sup><a>[[1]](#footnote-1)</a></sup>.
-
-Setting up a PNID on the 3DS is the same as setting up a NNID. You may either create the PNID on your console, or register from an account [on our website](/account/register) and link it to your console once you're ready.
-
-It is recommended to register the PNID on your device at this time, as registering on the website does not currently allow you to change your user data.
+On your system, go to **System Settings** and click on **Nintendo Network ID Settings**. From there, follow the instructions on screen to create a new Pretendo Network ID or link an existing one created [on our website](/account/register).
 
 <div class="tip red">
 	<strong>CAUTION:</strong>
-	A Pretendo Network ID may not use the same username as the account already linked to your 3DS! Ensure that you have a choose a different name for your PNID than the name on your NNID.
+	A Pretendo Network ID may not use the same username as the account already linked to your 3DS! Ensure that you have chosen a different name for your PNID than the name on your NNID.
 </div>
+
+<div class="tip">
+	<strong>There is a <a href="https://github.com/PretendoNetwork/nimbus/issues/32">known issue</a> where System Settings may still show your old NNID</strong>
+	<br>
+	A temporary workaround to fix this is to re-toggle between Nintendo Network and Pretendo Network via Nimbus.
+</div>
+
+It is recommended to create your PNID on your device at this time, as registering on the website does not currently allow you to change your user data.
+
+The 3DS does not rely on NNIDs for the vast majority of it's game servers. Because of this, using a PNID is also not required for most games<sup><a>[[1]](#footnote-1)</a></sup>.
 
 
 ## Other information


angeloanan:
On a future note, consider moving the recommendation of creating a PNID account to the site's registration page. One might simply miss the note by registering an account first before performing the installation guide.

tarikalmashni:
I highly support these changes and wish I knew about this issue when I created my PNID.
Obviously I used the username that I use everywhere for ages so I ran into this issue with currently no way for me to use Pretendo unless I create another account which definitely can't be an ideal solution.

Since the username should only be a field in the Pretendo database, imho it should be editable from the website's account management page. Looking at the Discord messages of users, this would prevent many of them as they would be able to fix this themselves.

jonbarrow:
> Since the username should only be a field in the Pretendo database, imho it should be editable from the website's account management page

Due to technical limitations on how some games/services work this just isn't possible. Nintendo also did not allow for username changes on Nintendo Network, likely for this same reason. You can blame Nintendo if you want, I guess

jonbarrow:
> It actually includes the step on how to create or link a Pretendo account **IN** the console

This seems unnecessary to me. It's the exact same process as adding a NNID on NN, telling the user exactly how to do that feels like a pretty extreme level of handholding

If one of the goals is to "be less wordy", then telling the user obvious information like this goes against that

docs/en_US/install/3ds.md
```diff
@@ -83,19 +83,25 @@ Depending on how you installed Nimbus, launch it either through the Homebrew Lau
 
 Your selection persists between reboots.
 
-## Signing into your PNID
+## Signing in into Pretendo Network
```
jonbarrow, line 86: This does not flow well, and I have never seen another service phrase this as "signing in into" before? Though you could argue that `into` in the current form is also incorrect

docs/en_US/install/3ds.md
```diff
@@ -83,19 +83,25 @@ Depending on how you installed Nimbus, launch it either through the Homebrew Lau
...
 
-The 3DS does not rely on NNIDs for the vast majority of it's game servers. Because of this, using a PNID is also not required for most games<sup><a>[[1]](#footnote-1)</a></sup>.
-
-Setting up a PNID on the 3DS is the same as setting up a NNID. You may either create the PNID on your console, or register from an account [on our website](/account/register) and link it to your console once you're ready.
-
-It is recommended to register the PNID on your device at this time, as registering on the website does not currently allow you to change your user data.
+On your system, go to **System Settings** and click on **Nintendo Network ID Settings**. From there, follow the instructions on screen to create a new Pretendo Network ID or link an existing one created [on our website](/account/register).
```
jonbarrow, line 88: One of your stated goals was to make these sections less wordy, but by adding this section you seem to have gone against that? This originally just informed users the setup process was exactly the same as a NNID, which told them everything they needed to know. Restating that here seems to just be repetitive for no valuable reason?

Also since the 3DS does not require a network account for basically all titles, the fact that you do not need a PNID was intentionally mentioned *first*, so that users could skip these steps if need be

All this has done is reorder the steps to make it somewhat less intuitive imo

angeloanan:
What seems to be obvious to you might not be obvious to a layperson who just wants to have their console's online functionality back (I know I'm one). It is not everyday that you go to the said section in the settings app, let alone creating / linking an NNID on to a 3DS. 

I know I'm biased; I've spent more than 10 minutes trying to figure out where to link my PNID after toggling Nimbus to Pretendo, and an extra 5 minutes scrolling through Nimbus' GitHub repo just to figure out that there was a known bug. But, in my opinion, a guide like this should be foolproof enough to the point where someone who doesn't understand a thing about 3DS hacking could follow it.

Also, I believe that you've misunderstood what a wordy sentence is; Adding extra information does not make something wordy :/

docs/en_US/install/3ds.md
angeloanan: I'm up to whichever.

My rational with this specific change is that the term `PNID` has never been introduced in this guide and one must infer `PNID` = `Pretendo Network ID`

docs/en_US/install/3ds.md
angeloanan: While it is mentioned that it is the same steps as setting up NNID, passage lacks information on **where** to create or link a NNID on the console. As creating an NNID is a once-and-forget thing, I believe that adding the location on where to do this helps a lay person.

> Also since the 3DS does not require a network account for basically all titles, the fact that you do not need a PNID was intentionally mentioned first, so that users could skip these steps if need be

I would argue that the information provided is not clear enough as the structure of the text conveys that it is the next step after using Nimbus. Putting this step in a separate "extras" section or clearly marking the step as optional in the heading might solve this.

jonbarrow:
> Also, I believe that you've misunderstood what a wordy sentence is; Adding extra information does not make something wordy :/

Adding extra information does make something more wordy if the added information is not of substance, which is where we have differing opinions. You could expand what you've added to also include detailed steps for every step in the linking process on the console. That would be added information, but would add no real value to the guide as it's already obvious and thus becomes "wordy"

I understand that you personally may have not known, and I understand that something obvious to one person may not be obvious to another, but there has to be a line where we can reasonably assume the user can take it from there. Trying to make any guide 10000% idiot proof will never work, you'll never be able to cover all the bases and if you do then the guide becomes unreadable due to length (I'm not saying that's the case here currently)

I'm not convinced that that is a big enough issue to warrant expanding upon in the guide? So far you are the only person I've personally seen who couldn't figure out how to link a NNID to their console. Based on some rough checks of our own data, there are 107,725 registered 3DS users, and of those users it appears that 105,742 of them have PNIDs linked. So unless my DB queries were wrong, that's less than 2,000 people without a PNID linked to their 3DS and of them it's not clear if that's by choice or not. Given that data, it doesn't seem like this is really an issue worth "fixing"? We can't always try to account for the lowest common denominator, there has to be a line

I also think you may have taken this review the wrong way. I am always open to changes, but only when they make sense to actually change. If proposed changes don't initially seem to solve a real problem, are mostly someones personal preferences, make little to no meaningful changes, etc. then I *do* typically question them (again, not saying that's what happened here, but it has happened in other PRs). I firmly believe it's beneficial to everyone involved, both us and contributors, if changes are justified. So I hope the review didn't discourage you from continuing to make contributions (though we will be updating our contributing guide soon to help with this)

I am aware that I am not infallible though, and that my views and opinions may not actually represent the best needs for the community as a whole. Which is why I request review from and tag other dev team members for input. If I'm out voted I typically don't fight back on it, and accept the majority opinion

I do hope this comes off in the respectful way I intend it to. I know sometimes I have a habit of coming off as cold or standoffish over text, but I mean no harm :+1: I'm simply opening a dialog

ExperiencersInternational:
Looking at this, I feel like Jon is completely in the right here, seems unnecessary rewriting it when it already contains the necessary information (me and Jon helped overhaul the 3DS page in the first place).

It's fairly simple to do now, there's nothing more than needs to be added imo.

docs/en_US/install/3ds.md
```diff
@@ -83,19 +83,25 @@ Depending on how you installed Nimbus, launch it either through the Homebrew Lau
...
 
 <div class="tip red">
 	<strong>CAUTION:</strong>
-	A Pretendo Network ID may not use the same username as the account already linked to your 3DS! Ensure that you have a choose a different name for your PNID than the name on your NNID.
+	A Pretendo Network ID may not use the same username as the account already linked to your 3DS! Ensure that you have chosen a different name for your PNID than the name on your NNID.
```
ashquarky, line 92: [A user on Discord](https://discord.com/channels/408718485913468928/1067162922989662360/threads/1237976845962576016) suggested the following wording:

> CAUTION: A Pretendo Network ID cannot use the same username as the NNID already linked to your 3DS! Ensure that you choose a different username for your PNID than the username on your NNID.

They said "may not" was vague, which is fair, since "may" often means "optional". "Must not" would be another option. Note how they also said "username" to differentiate from e.g. Mii name.

PR #289, opened 2024-04-24
docs(install/3ds): Improve preamble section

angeloanan:
This PR aims to clean up the preamble section of the 3DS guide.

Specifically, it inlines the multiple cautions and tips by turning it into a dedicated section, as well as extending the guide's introductory statement by mentioning what the guide will do.

CleanShot 2024-04-24 at 12 10 45

## Detailed changes

* Inlined multiple tips and caution to be a dedication section "Things to know before you start" section
* Introduced an opening statement on how to connect to pretendo and what the guide will achieve
* Extended the guide steps to add the "Signing into your PNID" section

### DIFF
--- a/docs/en_US/install/3ds.md
+++ b/docs/en_US/install/3ds.md
@@ -1,23 +1,24 @@
 # 3DS/2DS Family
 
-<div class="tip red">
-	<strong>CAUTION:</strong>
-	SYSTEM TRANSFERS ARE NOT CURRENTLY SUPPORTED BY OUR SERVERS. ATTEMPTING TO PERFORM A SYSTEM TRANSFER MAY PREVENT YOU FROM BEING ABLE TO GO ONLINE IN THE FUTURE. SUPPORT FOR SYSTEM TRANSFERS IS IN DEVELOPMENT.
-</div>
+## Things to know before you start
 
-<div class="tip red">
-	<strong>CAUTION:</strong>
-	Collecting badges in Nintendo Badge Arcade while connected to one network and then launching the game on a different network will result in your badges disappearing. This occurs because the locally saved data does not match the data stored on the server.
-</div>
+This guide assumes that you have a **Homebrewed System** running the latest version of Luma3DS (13+). If you are unsure or do not have a homebrewed system, please follow the [3DS Hacks Guide](https://3ds.hacks.guide/) to check, install and update it. 
 
-<div class="tip">
-	ℹ️ This guide assumes that you have a <b>Homebrewed System running the latest version of Luma3DS (13+)</b>, if you don't please follow this <a href="https://3ds.hacks.guide/" target="_blank">guide</a> on how to homebrew your system first.
-</div>
+**Support for system transfer is currently in development**. Attempting to perform one while being connected to Pretendo will prevent you from being able to go online in the future. If you need to do a system transfer, please do it before continuing with this guide.
 
-The following steps are required for you to connect to the Pretendo Network:
+Badges in **Nintendo Badge Arcade** might disappear and become de-synchronized when switching between Pretendo and Nintendo Network. This is due to your local data not matching the data stored on the respective server.
+
+## Getting Started
+
+To connect to the Pretendo Network, you will need to install a homebrew application and enable Luma patches on your console.
+
+Nimbus is our homebrew application that allows you to connect and switch from Nintendo Network to the Pretendo Network.
+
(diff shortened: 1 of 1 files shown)


docs/en_US/install/3ds.md
```diff
@@ -1,23 +1,24 @@
...
-	<strong>CAUTION:</strong>
-	SYSTEM TRANSFERS ARE NOT CURRENTLY SUPPORTED BY OUR SERVERS. ATTEMPTING TO PERFORM A SYSTEM TRANSFER MAY PREVENT YOU FROM BEING ABLE TO GO ONLINE IN THE FUTURE. SUPPORT FOR SYSTEM TRANSFERS IS IN DEVELOPMENT.
-</div>
+## Things to know before you start
 
-<div class="tip red">
-	<strong>CAUTION:</strong>
-	Collecting badges in Nintendo Badge Arcade while connected to one network and then launching the game on a different network will result in your badges disappearing. This occurs because the locally saved data does not match the data stored on the server.
-</div>
+This guide assumes that you have a **Homebrewed System** running the latest version of Luma3DS (13+). If you are unsure or do not have a homebrewed system, please follow the [3DS Hacks Guide](https://3ds.hacks.guide/) to check, install and update it.
```
jonbarrow, line 5: I'm not sure I agree with removing the caution callouts. This was done intentionally to be eye grabbing and bring attention to the information inside. These are pretty important things to know about, and having them just look like the rest of the page makes it more likely someone will skim over them

@hauntii @gitlimes @DaniElectra thoughts?

docs/en_US/install/3ds.md
```diff
@@ -1,23 +1,24 @@
...
 
-<div class="tip red">
-	<strong>CAUTION:</strong>
-	Collecting badges in Nintendo Badge Arcade while connected to one network and then launching the game on a different network will result in your badges disappearing. This occurs because the locally saved data does not match the data stored on the server.
-</div>
+This guide assumes that you have a **Homebrewed System** running the latest version of Luma3DS (13+). If you are unsure or do not have a homebrewed system, please follow the [3DS Hacks Guide](https://3ds.hacks.guide/) to check, install and update it. 
 
-<div class="tip">
-	ℹ️ This guide assumes that you have a <b>Homebrewed System running the latest version of Luma3DS (13+)</b>, if you don't please follow this <a href="https://3ds.hacks.guide/" target="_blank">guide</a> on how to homebrew your system first.
-</div>
```
jonbarrow, line 15: Removing this as a callout makes it inconsistent with the Wii U guide

jonbarrow (commented):
What was the purpose of removing the callouts? The information hasn't changed, you've simply removed the eye grabbing nature of them which was intentional. I don't see those as valuable changes, and could actually be worse for the experience as it makes it more likely that readers would skim past important information

docs/en_US/install/3ds.md
DaniElectra: Yeah I disagree with inlining the cautions and tips, I think how they are rn it's already fine

angeloanan: I would argue that having 2 back-to-back callouts in **the beginning** of the guide is almost too much. A person who just wants to get Pretendo installed might ignore the callouts and skip straight to the body of the guide.

I think that having a "Prerequisites" section would be better as it is the first thing that they will read anyway.

angeloanan:
> ... and could actually be worse for the experience as it makes it more likely that readers would skim past important information

I would argue that the reverse thing would be true. It is not intuitive to have an all-caps back-to-back callout on the very top of the screen, especially when the same information could just be conveyed as a normal section of a guide that the user firstly read.

docs/en_US/install/3ds.md
angeloanan: I'm fine with whichever.

Will make the change in a bit if necessary!

docs/en_US/install/3ds.md
jonbarrow: Having callouts for important information at the start of guides/wikis is fairly standard practice. You can see this on wikis like Wikipedia and even the official homebrew guide. It's done this way for a reason, to be immediately eye catching to the user. Putting key information like this as plain text like anything else definitely runs more of a risk of people missing it. Yes, people can still skip callouts. But it's far less likely

Screenshot from 2024-04-24 11-09-07
Screenshot from 2024-04-24 11-09-25

PR #292, opened 2024-05-01
Make error support pages more dynamic and use all known error codes

jonbarrow:
Currently our support pages are fairly lacking. We only have a page for a single error not related to Miiverse and not for one of our custom errors. The way the site is structured also makes it difficult to manage translations since all the pages are markdown.

This PR addresses a few issues:

1. Makes the website use all known error codes. Currently Bandwidth knows about far more errors than the website, but provides far less information than the website would
2. All error code information is now handled as JSON, and uses a single markdown template. This makes it possible to integrate these errors into translation tools and keep the pages looking uniform
3. Transitions away from using the term `Search` when we mean `Error`. Right now the way its structured the "search" page implies site-wide searching at a glance, only to show users it only searches errors. This can be confusing, so this PR also transitions to just calling everything what it is; errors. We can re-add a search page back at a different time if we want to introduce site-wide search

The error code information now comes from https://github.com/PretendoNetwork/error-codes. This data was mostly generated from Bandwidth's error code list (which will also be deprecated to use this same repo).


jonbarrow:
Anyone have any objections? Everything is ready to merge now

DaniElectra (commented):
Ideally we'd also want to show the module name of an error code, maybe above or below the `Applies to` section?

src/routes/docs.js
```diff
@@ -24,14 +27,22 @@ router.get('/install', async (request, response) => {
 	response.render('docs/install', renderData);
 });
 
-router.get('/search', async (request, response) => {
+router.get([
+	'/search', // TODO - Deprecate search page
```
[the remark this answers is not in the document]
jonbarrow: Existing links out in the wild, mostly

src/routes/docs.js
```diff
@@ -50,6 +61,42 @@ router.get('/:page', async (request, response, next) => {
 	response.render('docs/docs', renderData);
 });
 
+router.get([
+	'/errors/:errorCode', // TODO - Deprecate "errors" route
```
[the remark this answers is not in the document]
jonbarrow: Existing links in places like Discord

jonbarrow:
> maybe above or below the Applies to section?

That works! I'll see what looks nice

src/routes/docs.js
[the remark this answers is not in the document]
DaniElectra: that is fair 👍

src/routes/docs.js
```diff
@@ -24,14 +27,22 @@ router.get('/install', async (request, response) => {
...
 	const renderData = {
 		errorList: JSON.stringify(errorList),
-		currentPage: 'search',
+		currentPage: 'errors',
 	};
+
 	response.render('docs/search', renderData);
 });
 
+router.get('/error', async (_request, response) => {
```
wolfendale, line 42: Any particular reason for adding this? I'd be wary especially of adding a 301 as that will get cached by browsers and potentially be an issue if you ever wanted to add a page at this path in the future.

I'd recommend using a 303 if you really want to do the redirect

src/routes/docs.js
```diff
@@ -50,6 +61,42 @@ router.get('/:page', async (request, response, next) => {
...
+	const locale = response.locals.localeString;
+	const error = request.params.errorCode;
+
+	const [ sysmodule, errorCode ] = error.split('-');
+
+	if (!errorCode) {
+		return next();
+	}
+
+	let template = fs.readFileSync(path.join(__dirname, '../../docs/common/error-page-template.md'), {
```
wolfendale, line 77: I'd recommend doing something like:
```js
let template = await fs..promises(path.join(__dirname, '../../docs/common/error-page-template.md'), 'utf8');
```
So that you're not blocking the node thread each time it needs to respond to a request to this page.

docs/common/error-page-template.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Message
+
+> {message}
+
+## Cause
+
+{description}
+
+### Solution
```
gitlimes, line 14: Is this supposed to be an h3?

gitlimes: the solution specifically, dunno why it selected four rows, I'm on mobile

src/routes/docs.js
```diff
@@ -50,6 +61,42 @@ router.get('/:page', async (request, response, next) => {
...
+		encoding: 'utf8'
+	});
+
+	const errorInfo = errors.getErrorInfo(sysmodule, errorCode, locale);
+
+	if (!errorInfo) {
+		return next();
+	}
+
+	template = template.replace('{module}', sysmodule);
```
wolfendale, line 87: This will iterate the whole template string for each replace and will only replace a single instance of each of the placeholders

I'd recommend taking a look at the [express markdown example](https://github.com/expressjs/express/blob/master/examples/markdown/index.js#L17-L25) where they perform a single replace and use a function to look up the right data. Something like this could be added to the `util.parseDoc` maybe?

docs/common/error-page-template.md
jonbarrow: No, good catch :+1:

src/routes/docs.js
jonbarrow: No reason in particular. Just a little safety net for people who mistype the URL, since it's clear what page they'd actually want to go to (the search path is `/docs/errors`, and the support pages are at `/docs/error/{code}`. I can easily see someone just removing `{code}` and ending up at `/docs/error` but meant `/docs/errors`)

Since we have a `/docs/errors` page I don't see us making a real `/docs/error` page in the future, to avoid confusion. Though if we were to change the status code 307 may be more appropriate?

While 303 *can* be used for `GET`, it's most typically used for `POST` requests. 303 is also intended to indicate that the server does not have a representation for the requested resource, and suggests one that *may* represent it. But the server can't be sure that the resource being redirected to actually represents the one the user wants

307 on the other hand assumes to know what resource the client wants (which I would argue is the case here), and redirects them. 307 also tells the client to NOT cache the redirect and still use the original target, which sounds like a better fit than 303?

Honestly though there's not a perfect fit for a status code here if you want to reserve the path for future use

src/routes/docs.js
jonbarrow: While this won't block the node thread, this increases the read time by quite a bit. Even the slowest instance of the sync version was faster than the fastest instance of the async version

Using the `promises` version of `readFile` in my testing takes between 0.331ms-1.141ms. Whereas the sync version takes between 0.059ms-0.168ms

Though at these speeds both are basically instant, so that may not matter much, but the sync version is so fast that the block is *also* essentially unnoticeable

src/routes/docs.js
jonbarrow: Correct, it will iterate the whole string. However it's an very small string, and the operations are extremely fast

Because of this I prioritized readability over trying to micro-optimize these operations

However to give some numbers:

On a test of 200 iterations the string-replace method took on average 0.016665ms to complete, between 0.008ms and 0.143ms. Using the same kind of regex-replace method as shown in the example you linked (which was what I initially did use), is actually slower in this case. In the same sample of 200 iterations it took an average of 0.02626ms to complete, between 0.013ms and 0.265ms

src/routes/docs.js
wolfendale: I think 303 or 307 are both fine, the main technical difference is that 303 forces the redirected request to a GET, but 307 doesn't. In this case you're just using it on a GET endpoint anyway, sound they're functionally equivalent. I don't really have a preference between the two but think it's worth not letting the browser cache the redirect :)

src/routes/docs.js
wolfendale: It's a tiny bit slower than sync, but it's being not time taken up on the node thread, so although it'll make each individual request a _tiny bit slower_ it means you can respond to more requests at once, either way it's probably not a massive difference. I'm just used to high traffic systems where we optimise for overall throughput above individual request times

jonbarrow: I would normally agree but even the response time was essentially the same. We use morgan to log response timings and in all my tests with both versions there was essentially no difference in response time

Also for context our site is relatively high traffic. In the past 7 days we've had 1.82m total requests (237.37k being in the past 24 hours) from 55.45k unique visitors. But we also use Cloudflare which handles a lot of caching for pages like this, so I'm not SUPER concerned about these kinds of micro optimizations right now

If it presents an issue later it can be addressed definitely, but as is there's no issues

src/routes/docs.js
wolfendale: > Correct, it will iterate the whole string. However it's an very small string, and the operations are extremely fast
Totally agree, was more pointing it out rather than saying it wouldn't be performant

I think my main point was that it'll only replace a single instance of each of the placeholders. This totally works for the current template, but if you ever added the same placeholder in another spot in the markdown that would require you update the code here too. I appreciate that might be incredibly unlikey and a good case for YAGNI, but thought I'd mention it 👍

jonbarrow: Of course, even if we ultimately disagree with a suggestion please do feel free to still make them :+1: I value all feedback and suggestions, even ones we end up not going with

Breaking this out into a more generic system is not a *bad* idea, and in fact may be something to consider again if we were to add these kinds of translation templates for other pages (such as the docs pages). But currently yes this is a good case for YAGNI, we only need this kind of template system for one page and this was the simplest/fastest way to get that done

Though I did misunderstand your suggestion originally. I assumed you made it for performance reasons, due to your previous comment and because you brought up to amount of passes through the string, so that was one me

Also it should be noted that the current state of the website is, for all intents and purposes, temporary. @hauntii is working on a total rewrite of the website, migrating it to TypeScript and using Next rather than express+handlebars. So all of this will be trashed eventually anyway

src/routes/docs.js
jonbarrow: I decided to go with 307, purely for semantic reasons (I kinda sort of be a stickler there, it's a flaw I know) :+1:

jonbarrow:
The solution heading size has been fixed and the module information added. Should be good now

Screenshot from 2024-05-04 12-47-11

gitlimes (approved):
LG TV!

PR #295, opened 2024-05-07
[Feature]: Discourse SSO

jonbarrow:
As discussed in https://github.com/PretendoNetwork/Pretendo/pull/64, we have an accessibility issue when it comes to information and support. Part of this comes down to our reliance on Discord for a lot of things, which we have been criticized for in the past (and rightfully so).

This PR aims to rectify that. This PR adds in support for [DiscourseConnect](https://meta.discourse.org/t/setup-discourseconnect-official-single-sign-on-for-discourse-sso/13045) SSO. Once this PR is merged, and our Discourse forum hosted, we will be able to open up our forums to users utilizing our PNID account system.

Having a dedicated, web-based, forum will help a lot. We will be able to reach people outside of Discord, including those who don't want to use the platform or are unable to access it/our server (for example banned users, who even if banned from our Discord are still entitled to PNID support).

This allow will allow us, again as discussed in https://github.com/PretendoNetwork/Pretendo/pull/64, to have meaningful technical discussions with contributors outside of our developer channels on Discord (which are currently a supporter perk, one which I am hesitant to make public out of fear of taking something away from supporters).

I am still working out some kinks with Discourse, but their support team has been top notch so that shouldn't take long.

PR does include some hacks, but that's because the website is written poorly in general. The rewrite will clean it up a lot.

### DIFF
--- a/src/routes/account.js
+++ b/src/routes/account.js
@@ -71,7 +71,8 @@ router.get('/', requireLoginMiddleware, async (request, response) => {
 
 router.get('/login', async (request, response) => {
 	const renderData = {
-		error: request.cookies.error_message
+		error: request.cookies.error_message,
+		loginPath: '/account/login'
 	};
 
 	response.render('account/login', renderData);
@@ -88,7 +89,6 @@ router.post('/login', async (request, response) => {
 		response.cookie('token_type', tokens.token_type, { domain: '.pretendo.network' });
 
 		response.redirect(request.redirect || '/account');
-
 	} catch (error) {
 		console.log(error);
 		response.cookie('error_message', error.message, { domain: '.pretendo.network' });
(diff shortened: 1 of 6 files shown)


src/routes/account.js
```diff
@@ -407,5 +407,138 @@ router.post('/stripe/webhook', async (request, response) => {
...
+
+			// * Discourse REQUIRES unique emails, however we do not due to NN also
+			// * not requiring unique email addresses. Email addresses, for now,
+			// * are faked using the users PID. This will essentially disable email
+			// * for the forum, but it's a bullet we have to bite for right now.
+			// TODO - We can run our own SMTP server which maps fake emails (pid@pretendo.whatever) to users real emails
+			const payload = Buffer.from(new URLSearchParams({
+				nonce: decodedPayload.get('nonce'),
+				external_id: accountData.pid,
+				email: `${accountData.pid}@invalid.com`, // * Hack to get unique emails
```
ashquarky, line 440: If we do ever do `<pid>@pnid.pretendo.network` (or w/e), is it going to be a pain to migrate the existing accounts tied to invalid.com addresses?
Could this be a security risk if the owner of invalid.com were to start requesting password resets?

jonbarrow: These invalid emails only exist for the Discourse user profiles, and we can mass update those once we have an SMTP server.

There is no password resets because this uses DiscourseConnect, which replaces the original Discourse login flow with one from an external service (in this case, PNID login on the website).

Using `whatever@invalid.com` was the official recommendation from Discourse to handle this situation ([relevant comment](https://meta.discourse.org/t/integration-into-custom-auth-system-where-emails-are-not-unique/306489/28?u=jonbarrow)).

ashquarky: Oh wow, I didn't realise how in-depth they went with you. LGTM then

gitlimes (approved):
looks good to me!

PR #297, opened 2024-05-16
update documentation that is innacurate

treatmeniceplz:


gitlimes (changes requested):
Hi @treatmeniceplz.
Thanks for your contribution to the project!

After reviewing your proposed change, I believe that the change to the link might actually be detrimental: the page you have linked to does not provide information on how to set up CEMU for online play, while the original did. This could have the potential outcome of confusing users who are unfamiliar with the emulator.

@jonbarrow, if you agree, could you please close this PR? Otherwise, feel free to share your thoughts.

Thank you for your cooperation.

jonbarrow:
I agree with @gitlimes. The original link is specifically about setting up Cemu for online play, and on that page it has a "requirements" section mentioning the use of Dumpling already.

Changing our documentation to only point to the page regarding Dumpling would only confuse people.

PR #300, opened 2024-05-21
Fix upgrade.js script tag

MatthewL246:
On the upgrade page, the `upgrade.js` script uses a self-closing `<script>` tag. This is not valid HTML.

As far as I can tell, in production, the upgrade script is only loaded because Cloudflare adds an analytics script, which adds an extra `</script>` tag to the end of the HTML. This leads to Chrome parsing the page like this, which doesn't seem right:
image


gitlimes:
Oh wow, I absolutely messed up... Thank you for fixing it, that's a great catch!

Issue #279, opened 2024-04-15
BUG: Japanese account menu text broken (mobile)


reindex-ot:
layout is broken in Japanese.
![Screenshot_2024-04-15-20-49-57-582_com android chrome](https://github.com/PretendoNetwork/website/assets/32851879/93b2830a-eafd-4928-b7ce-4b7700f1c398)
![Screenshot_2024-04-15-20-49-47-795_com android chrome](https://github.com/PretendoNetwork/website/assets/32851879/ae4ebdc8-6c0c-4679-a225-c1b630290fc1)

Silver-Volt4:
Should be fixed when dev is merged into main

reindex-ot:
nice.

gitlimes:
Yup, this issue has been fixed, the patch just isn't live yet. Closing this

PR #301, opened 2024-05-23
Fix CJK characters in user widget

Silver-Volt4:
Adds a CSS rule to prevent CJK chars from pointlessly breaking.
image
Fixes issue #279.


gitlimes:
Sweet! Thanks for your contribution!

Issue #299, opened 2024-05-20
Website allows people to spam emails to anyone


yoSauxy:
Hello there, I have recently been SPAMMED with emails because you can send a password recovery email to anyone's email just by entering their PNID, I feel like this should 100% change
![Screenshot 2024-05-20 192815](https://github.com/PretendoNetwork/website/assets/48485396/59a36cf0-5290-484a-a10a-2fdebb4278a2)

jonbarrow:
This is how all username based account systems with a password reset work. It's not unique to this. Also it's impossible to remove PNIDs as input given that emails are not unique to specific users. Nintendo Network was designed to allow an email address to be reused by any number of accounts, and because of this so do we. Without using a PNID as input it's not possible to pull the users email address.

DanilochTop:
I was reading this issue and perhaps the possibility of adding a rate limit for sending emails based on the user's PNID could be taken into consideration.
This, in addition to preventing spam, would avoid excessive spending on email APIs.

jonbarrow:
Sure, but that's a different issue than the one described here.

PR #306, opened 2024-05-31
Add captcha to Forgot Password requests

DanilochTop:
Motivation: #299 

This PR aims to add captcha to forgot password requests on the front end, and do some changes on the back end to support this feature.

### Detailed changes
- Added hcaptcha box to the forgot password page.
- Changed the forgot password page to work with forms instead of using JS to handle inputs, so we can pass the hCaptchaResponse automatically.
- Added success and error message boxes instead of alert.
- Added cookie support to form input, so the user don't need to input the credentials again.
- Created a forgotPassword function in util module to handle requests/responses in a better way.

I also created a PR in the account server repository to support captcha, because it's verified there.

Any bugs, feedback or suggestions just message me! 😄


src/routes/account.js
```diff
@@ -152,12 +152,37 @@ router.get('/logout', async (_request, response) => {
 });
 
 router.get('/forgot-password', async (request, response) => {
-	response.render('account/forgot-password');
+    const renderData = {
```
MatthewL246, line 155: I noticed that there are inconsistent tabs and spaces in this file. The Pretendo projects use tabs for indentation as a standard.

DanilochTop: Thanks for showing me that! Something happened while I was commiting and the indentation got messed up.

MatthewL246: That commit didn't fix the issue.

DanilochTop:
Everything should be fine and ready to merge now.

jonbarrow:
LGTM

PR #311, opened 2024-06-20
Spanish translation

MLDvp:
I have started to translate the guide into spanish.


MLDvp:
since the guide was untranslated, i started to do it there, and unfortunately it wasn't in the locales folders...

Issue #318, opened 2024-07-28
[Enhancement]: Move FAQ to dedicated pages


jonbarrow:
### Checked Existing

- [X] I have checked the repository for duplicate issues.

### What enhancement would you like to see?

Currently our FAQ is "fine", but it's very small and slapped onto the home page. Building off of #317, which is going to require a much longer FAQ answer, it might be a good idea to consider making dedicated FAQ pages. Such as:

- Make https://pretendo.network/faq list the questions, with links to pages
- Make https://pretendo.network/faq/NAME_OR_ID_OR_SOMETHING actually show the long-form answer
- Maybe we can keep the short answers on the home page too and just link to the longer ones there?

PR #327, opened 2024-08-19
Add moderation FAQs, update team, minor wording tweaks

ashquarky:
Resolves #XXX

### Changes:
A few text changes on the homepage:
 - Added some FAQs regarding cheating online leading to bans. Apparently a lot of 3DS players think they "can't be caught" since bypassing bans were trivial on NN. Come upfront and say that no, it doesn't work like that here. Will hopefully discourage opportunistic cheating.
 - Add an FAQ entry to try and drive home "no eta for games". Point out that working on stabilisation is more important.
 - Add new team members - Matthew, Wolfi, and Trace. I have definitely forgotten someone but this keeps the grid layout even :blue_heart: 
 - Add major outside contributors (binaryoverload, jvs, simon, oatmealdome) to special thanks. Need to check if they actually want to be on here, but all of them have dev roles on the Discord so it's hardly a secret.

Changes are made to en_US locale so it'll probably take a while to filter into the others.

### DIFF
--- a/locales/en_US.json
+++ b/locales/en_US.json
@@ -62,12 +62,12 @@
 				"answer": "No. Lots of Pretendo's features/services are developed independently (for example, Miiverse may be worked on by one developer while Accounts and Friends is being worked on by another) and therefore we cannot give an overall ETA for how long this will take."
 			},
 			{
-				"question": "Does Pretendo work on Cemu/emulators?",
-				"answer": "Pretendo supports any client that can interact with Nintendo Network. Currently the only emulator with this kind of functionality is Cemu. Cemu 2.0 officially supports Pretendo under your network account options in the emulator. For information on how to get started with Cemu, check out the <a href='https://pretendo.network/docs/install/cemu'>documentation</a>.<br>Citra does not support true online play and thus does not work with Pretendo, and does not show signs of supporting true online play at all. Mikage, a 3DS emulator for mobile devices, may provide support in the future though this is far from certain."
+				"question": "When will you add more games?",
+				"answer": "We work on new games once we feel that our backend libraries are ready to support it, and there is developer time available to maintain it. A lot of our work goes into stabilising and completing our existing games - we want to get the best experience possible in those before we move on to new titles. Since new work comes up all the time, we cannot make any estimate of when that would be."
 			},
 			{
-				"question": "If I am banned on Nintendo Network, will I stay banned when using Pretendo?",
-				"answer": "We will not have access to Nintendo Network's bans, and all users will not be banned on our service. However, we will have rules to follow when using the service and failing to follow these rules could result in a ban."
+				"question": "Does Pretendo work on Cemu/emulators?",
+				"answer": "Cemu 2.1 officially supports Pretendo under your network account options in the emulator. For information on how to get started with Cemu, check out the <a href='https://pretendo.network/docs/install/cemu'>documentation</a>.<br>Some 3DS emulators or forks might support us, but we do not have any official recommendation or setup instructions at this time. The final builds of Citra do not support Pretendo."
 			},
 			{
 				"question": "Will Pretendo support the Wii/Switch?",
@@ -76,6 +76,14 @@
 			{
 				"question": "Will I need hacks to connect?",
 				"answer": "For the best experience on consoles, you will need to hack your system - specifially Aroma for Wii U and Luma3DS for 3DS. However, on Wii U, the hackless SSSL method is also available with limited functionality. See our <a href='/docs/install'>setup instructions</a> for details."
+			},
+			{
+				"question": "If I am banned on Nintendo Network, will I stay banned when using Pretendo?",
+				"answer": "We will not have access to Nintendo Network's bans, and all users will not be banned on our service. However, we will have rules to follow when using the service and failing to follow these rules could result in a ban."
+			},
+			{
+				"question": "Can I use cheats or mods online with Pretendo?",
+				"answer": "Only in private matches - gaining an unfair advantage or distrupting the online experience with people who didn't consent (as in public matches) is a bannable offense. We regularly apply account and console bans to both Wii U and 3DS systems. Pretendo uses extra security measures that make traditional 'unban' methods like changing your serial number ineffective."
 			}
 		]
 	},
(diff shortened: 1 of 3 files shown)


ashquarky:
Forgot to mention, GitHub contributors is listed twice on purpose to keep the grid rows even. If there's someone else I can put there instead please let me know.

locales/en_US.json
```diff
@@ -154,6 +162,24 @@
 				"caption": "Web and server development",
 				"picture": "https://github.com/hauntii.png",
 				"github": "https://github.com/hauntii"
+			},
+			{
+				"name": "MatthewL246",
+				"caption": "Server architecture and community work",
```
MatthewL246, line 168: Personally, I'd say my role is closer to "DevOps" than "server architecture."

MatthewL246 (commented):
The FAQ updates look good to me.

locales/en_US.json
ashquarky: I put anything related to docker as server architecture. I dunno if we want to be more specific for userfacing docs like this

ashquarky: (Mario has "Systems research and server architecture"; Billy has "Preservationist and server architecture" - I think Billy's contributions revolve around CloudFlare usage and configuration so it's kinda vague)

jonbarrow: Billy's main contributions, historically, have revolved around preserving packet dumps. He's the main one who was getting us information at a large scale prior to our own mass collection. The CloudFlare stuff came later, and is now mostly what he does since packet dumps are kinda dead

jonbarrow:
Since the FAQ is being updated, do we maybe want to tackle this issue? https://github.com/PretendoNetwork/website/issues/318

locales/en_US.json
```diff
@@ -62,12 +62,12 @@
...
 			},
 			{
-				"question": "Does Pretendo work on Cemu/emulators?",
-				"answer": "Pretendo supports any client that can interact with Nintendo Network. Currently the only emulator with this kind of functionality is Cemu. Cemu 2.0 officially supports Pretendo under your network account options in the emulator. For information on how to get started with Cemu, check out the <a href='https://pretendo.network/docs/install/cemu'>documentation</a>.<br>Citra does not support true online play and thus does not work with Pretendo, and does not show signs of supporting true online play at all. Mikage, a 3DS emulator for mobile devices, may provide support in the future though this is far from certain."
+				"question": "When will you add more games?",
+				"answer": "We work on new games once we feel that our backend libraries are ready to support it, and there is developer interest in maintaining it. A lot of our work goes into stabilising and completing our existing games - we want to get the best experience possible in those before we move on to new titles. Since new work comes up all the time, we cannot make any estimate of when that would be."
```
jonbarrow, line 66: This can VERY easily be read as "Pretendo only wants to support games they think are cool", due to the wording of "and there is developer interest in maintaining it",which is not true.

I understand that *some* people on the team only wish to work on games they personally find interesting, but that is NOT the stance of the project as a whole, and there are plenty others who work on all games.

Our goal is to support all games within our scope, not just those which have developer interest.

locales/en_US.json
```diff
@@ -62,12 +62,12 @@
...
 			},
 			{
-				"question": "If I am banned on Nintendo Network, will I stay banned when using Pretendo?",
-				"answer": "We will not have access to Nintendo Network's bans, and all users will not be banned on our service. However, we will have rules to follow when using the service and failing to follow these rules could result in a ban."
+				"question": "Does Pretendo work on Cemu/emulators?",
+				"answer": "Cemu 2.0 officially supports Pretendo under your network account options in the emulator. For information on how to get started with Cemu, check out the <a href='https://pretendo.network/docs/install/cemu'>documentation</a>.<br>Citra does not support true online play and thus does not work with Pretendo, and does not show signs of supporting true online play at all. Mikage, a 3DS emulator for mobile devices, may provide support in the future though this is far from certain."
```
jonbarrow, line 70: Since this PR was made Cemu has released version 2.1. This answer was always a bit misleading since "Cemu 2.0" actually does NOT support us (Cemu 2.0 is the stable release, and we got support in Cemu 2.0-5 which is an experimental release). However Cemu has no released 2.1, which is the first stable release to support us

Also I wonder if we should give a nebulous "some 3DS emulators/forks may support us" type of deal. Pablo's fork does work on our servers, but there's concerns from his end about Nintendo getting trigger happy (which is understandable, and why he doesn't provide any setup instructions himself). Also Citra is dead so we probably shouldn't talk about it like it's still alive

locales/en_US.json
```diff
@@ -76,6 +76,14 @@
...
 				"question": "Will I need hacks to connect?",
 				"answer": "For the best experience on consoles, you will need to hack your system - specifially Aroma for Wii U and Luma3DS for 3DS. However, on Wii U, the hackless SSSL method is also available with limited functionality. See our <a href='/docs/install'>setup instructions</a> for details."
+			},
+			{
+				"question": "If I am banned on Nintendo Network, will I stay banned when using Pretendo?",
+				"answer": "We will not have access to Nintendo Network's bans, and all users will not be banned on our service. However, we will have rules to follow when using the service and failing to follow these rules could result in a ban."
+			},
+			{
+				"question": "Can I use cheats or mods online with Pretendo?",
+				"answer": "Only in private matches - cheating online with people who didn't consent (as in public matches) is a bannable offense. We regularly apply account and console bans to both Wii U and 3DS systems. Pretendo uses extra security measures that make traditional 'unban' methods like changing your serial number ineffective."
```
jonbarrow, line 86: Maybe this should be updated to note that "cheats" which do not give any advantage are also allowed, even in public matches? There have been questions about that. We don't have an "approved cheats list" or anything but common sense

jonbarrow (changes requested):

locales/en_US.json
ashquarky: I think the public documentation should be as strict as possible, then we can choose to be lenient - common sense - as appropriate. (Discord takes the same approach to 3rd-party clients, for example.)

Outright saying "no advantage is okay" can lead to weird edge cases. In Splatoon, is forcing a specific stage an advantage? Is night mode an advantage? They're certainly very disruptive to the online experience and we don't want people doing it, but documenting carveouts like this could corner us.

jonbarrow: > They're certainly very disruptive to the online experience

Then this would obviously not be allowed. The official stance we've always had on cheats/mods in public sessions is that they are not allowed if they:

- Give you an advantage
- Disrupt the game for other users

Cheats like the ones you described would easily fall into the 2nd category.

The issue I see right now is that there will be a discrepancy between the information on the website and what we have officially said elsewhere (not just Discord, but on basically all of our socials as well), which is that cheats/mods which don't harm the experience of others are allowed in public sessions. This is different from how people like Discord handle this because we have explicitly *allowed* cheats/mods which don't fall into those categories. This can easily result in users not understanding what we do or do not actually allow, which can turn into more confusion, questions, and false reports.

ashquarky: Honestly, with the way moderation has been going, I do wonder if changing our stance to be flat-out "no cheats"  - as per Discord - is a good option. Lots of people "accidentally" enable things or "forget it was on" or whatever. We could then have a small whitelist of definitely allowed things like Rival Unpatch for Splatoon or custom skins in Minecraft. That's a discussion to have in private though instead of on this FAQ update, so I'll adjust it as you suggested for the time being.

ashquarky:
> Since the FAQ is being updated, do we maybe want to tackle this issue? #318

I would prefer to just do text updates for now, I'm not confident in my ability to get the CSS right.

locales/en_US.json
ashquarky: "Developer time available" maybe?

jonbarrow (approved):

PR #363, opened 2024-12-12
Discourse SSO updates

### LINKED ISSUE #345: [Enhancement]: Add group syncing to Discourse SSO
### Checked Existing

- [X] I have checked the repository for duplicate issues.

It would be convenient to have groups on the forum that are automatically synced based on account state.

This would include:
- Developer (access level 3)
- Juxtaposition moderator (access level 2)
- Tester/supporter (access level 1)

It may also be possible to check if the PNID's linked Discord account has a moderator role in the server, which could be used to create a Discord moderator group.

DiscourseConnect has support for [adding and removing groups with SSO](https://meta.discourse.org/t/setup-discourseconnect-official-single-sign-on-for-discourse-sso/13045#p-41969-specifying-group-membership-7). It would probably be best to use the `add_groups` and `remove_groups` attributes instead of the `discourse connect overrides groups` setting, as that would break forum-only groups like category moderators.

### LINKED ISSUE #352: [Feature]: Allow users to use their Mii nickname on the forum
### Checked Existing

- [X] I have checked the repository for duplicate issues.

### What feature do you want to see added?

Basically, what I want is to allow users to display their Mii's nickname on the forum, alongside their PNID.

It would be similar to how Roblox DevForum handles it, where your display name is set as your name and your username on Roblox is set as your username on the forum.

Screenshot_20241106-202653

### Why do you want to have this feature?

Essentially just to reduce confusion. Some of us have long usernames and people don't always know what to call us.

Take me for example, I'm known as GamersInternatio on the forum and it can be confusing to figure out what a user should call me: Gamers, Gamer, or even Intern as I found out earlier. By allowing users to use their Mii nicknames on the forum, this will ease confusion as users will know that they're able to call me 'Flynn' instead.

Based on what I could gather from a quick check, this would literally only require a line change.

https://github.com/PretendoNetwork/website/commit/255e68a881d9d201adb761c4f7af7e1d6698bd7d

Currently `name = accountData.username` but in the future, we can use `name = accountData.mii_name` instead and enable `SiteSetting.auth_overrides_name`.

If you want any additional information, staff should be able to access this private DM exchange: https://forum.pretendo.network/t/moderation/4261/371?u=gamersinternatio

MatthewL246:
Resolves #345 and #352

### Changes

This adds new features to the Discourse SSO that will sync a user's groups based on their PNID's access level, Stripe tier, and linked Discord account's roles in the server. The access levels, tier levels, role IDs, and group names are all controlled by values in the config file for easy modification.

This also adds a new feature to manually sync a user's SSO state without requiring them to sign out and back in. It is currently used after handling a Stripe event, and it could be used in the future when a user is promoted/demoted as a moderator/developer. This would address the concerns shared in #345 over a user purposefully staying signed in to avoid being removed from a privileged group. For now, I think using this after handling a Stripe event and having a forum admin manually remove a user's groups when they are demoted will work fine, considering how rarely the second event happens.

Setting up the manual sync feature will require creating a Discourse API key. I would suggest setting its user to system, permissions to granular, and only enabling the sync SSO route for security reasons.

### Prerequisites

This requires PretendoNetwork/account#128 to be merged first.

### Current issues

As I commented in `createDiscoursePayload`, I noticed some issues with primary groups while testing this. I wouldn't consider them showstoppers because this basically only affects developers and moderators, none of a user's groups will change during most logins, and users can select their own primary group in their profile settings.

### Testing

I tested this as thoroughly as I could using self-hosted instances of the website (set up with a real Discord application and bot) and Discourse. The only issue is the Stripe integration, as I don't have a Stripe account. I tested the Stripe tiers functionality by creating fake Stripe tiers in the database under the assumption that they are just regular integers. I also manually tested the `discourseUserExists` and `syncDiscourseSso` functions used in the Stripe event handling by using the debug console. However, adding the SSO syncing to the end of 'handleStripeEvent` is a bit of an assumption and should be reviewed.

- [x] I have tested **most** of my changes.


MatthewL246:
I hope you consider rewording this message to specify that this CLA is
the Developer Certificate of Origin. My initial reaction to this
notification was negative because I worried that this was one of those
CLAs corporations use to give the project owners special rights over
others' contributions.

binaryoverload:
> I hope you consider rewording this message to specify that this CLA is the Developer Certificate of Origin. My initial reaction to this notification was negative because I worried that this was one of those CLAs corporations use to give the project owners special rights over others' contributions.

I hear your thoughts - and I agree the phrasing lends itself to worry people. Unfortunately, it's not something the CLA bot allows us to configure - so we'll just have to rely on people clicking on the CLA and seeing that it's just the certificate of origin

MatthewL246:
Nice!

PR #367, opened 2024-12-27
Reduce Mii editor HTTP requests from 4 to 1 on every change

ariankordi:
Currently on the Pretendo Mii editor, four requests are made every single time a change is made to the Mii. Here's what those are and how this PR addresses them:
* Two requests for icons meant for the save screen, representing the old Mii with a sorrow expression and new Mii with a happy expression.
  - Needlessly updated on every single change when they only show on the save screen.

Those icons are now deferred and not requested on every change.
* Two more requests for the actual Mii head shown and a bald version.

The bald head is meant to maintain the neck when a Mii's hair goes behind it. Here's what would happen if you overlaid the head of such a Mii without accounting for this:

But more requests to Nintendo is never a good thing, and they have a solution for this that had not been in the mii-js repo for the longest time: the `splitMode` param puts the front part and back part of the head in the same image. That is used by this commit and the result looks like it should.
  

(I held off on submitting the PR to the website back in September because I thought it needed a corresponding change to add the splitMode field in mii-js. Nobody responded to my PR. Turns out it this still works without that. OOOOOOOPS!)

### DIFF
--- a/public/assets/js/miieditor.js
+++ b/public/assets/js/miieditor.js
@@ -5,7 +5,7 @@
  *
  * browserify is needed for the use of require() in the browser
  */
-const Mii = require('@pretendonetwork/mii-js').default;
+const Mii = require('mii-js');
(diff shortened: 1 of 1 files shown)


ariankordi:
@binaryoverload Noticed that you ran CI on this earlier today. The eslint issues should be fixed now.

ariankordi:
Anyone interested in merging this?

gitlimes (approved):
LGTM!

PR #368, opened 2025-01-20
Update ESLint and Minor Tweaks/Fixes

### LINKED ISSUE #294: Redirection Not working for external Sites
For example I want to login into my account from the MK8 Website and the login for pretendo  website has the redirect parameter, however, after logging in it takes me to the home Page of my account and not the MK8 Website

binaryoverload:
Resolves #294 

### Changes:
- Move all references to `config.json` to a central `config.js` file for easy tracking of what config exists
- Move blog IO actions to use async IO for performance
- Stop /blog/:slug requests from requesting the list of posts for no reason
- Update eslint to version 9 and introduce common linting rules from @pretendonetwork/eslint-config
- Fix #294 by allowing https:// redirections for configurable domain suffixes 


public/assets/js/account.js
```diff
@@ -7,15 +7,15 @@ editSettingsModalButtonClose?.addEventListener('click', () => {
 	editSettingsModal.classList.add('hidden');
 });
 
-document.addEventListener('click', event => {
+document.addEventListener('click', (event) => {
```
jonbarrow, line 10: Was this done manually or as part of eslint? I'm a little confused on this one since in this file the `()` are being *added*, but later on in the PR there's files where the `()` are being *removed*:

Screenshot from 2025-01-21 12-18-14

This feels inconsistent?

src/routes/account.js
```diff
@@ -201,9 +201,9 @@ router.get('/connect/discord', requireLoginMiddleware, async (request, response)
 		tokens = await discordOAuth.tokenRequest({
 			code: request.query.code,
 			scope: 'identify guilds',
-			grantType: 'authorization_code',
+			grantType: 'authorization_code'
 		});
-	} catch (error) {
+	} catch (ignored) {
```
jonbarrow, line 206: What is this doing? I don't see `ignored` being used anywhere, I assume the name was changed to make it clear that the error isn't being used?

If that's the case, why not just remove it entirely? The `catch` block does not require the error to be captured:

```suggestion
	} catch {
```

jonbarrow (commented):
Overall looks good, just a couple questions. I had some other comments I was going to make but I largely see the website as temporary these days, just waiting to get rewritten

src/routes/account.js
binaryoverload: In JavaScript the catch parameters are required, I believe I tested this and removing it works in TS but not in pure JS

I can test again to be sure

public/assets/js/account.js
binaryoverload: This is all being done by eslint

I don't know the precise reason, but my guess is that single line expression lambdas don't have it, while multi line block lambdas do

src/routes/account.js
binaryoverload: Interesting - I must have been receiving the error before I set the ECMA version to latest - ig this is probably newish syntax

public/assets/js/account.js
binaryoverload: Just to follow up, this is the setting

It adds brackets as-needed (i.e. multiple parameters) but forces it for blocks

image

src/routes/account.js
jonbarrow: According to MDN the parameter was made optional after the initial spec for `try-catch`, yes. Most browsers implemented it in 2018, with the exception of Android's WebView which had it mid 2017 and then some which waited until 2020

So definitely recent-ish syntax. Idk when the spec was updated though

jonbarrow (approved):
LGTM :+1:

PR #369, opened 2025-01-28
Fix and optimise images

binaryoverload:
Resolves issues with images not loading

### Changes:
- Removes old broken jpg images
- Converts (most) images to webp, other than some PNGs which do not get smaller as webp
- Adds docs on how to convert to webp


public/assets/images/_README.md
```diff
@@ -0,0 +1,22 @@
...
+```bash
+mogrify -format webp *.png
+mogrify -format webp *.jpg
+```
+
+## Using cloudconvert
+
+If you don't want to install imagemagick, you can use cloudconvert to convert images to WebP.
+
+Go to https://cloudconvert.com/webp-converter and upload your images. You can convert multiple images at once.
```
jonbarrow, line 22: Cloudconvert limits the number of conversions you can do a day to 10 conversions unless you pay them. This limit applies to *all* of their tools simultaneously (so if you used 5 conversions in one tool, you now only have 5 left for every other tool)

As much as I love cloudconvert (I use them all the time) idk if we wanna shill for them like that?

binaryoverload: It's for this reason I listed imagemagick as most online sites have some sort of restrictions

Open to alternatives for a web converter

binaryoverload: https://towebp.io/ could be an option, doesn't appear to have limits

jonbarrow: I was in favor of just not having a cloud based option listed at all tbh, letting the user find that themselves if ImageMagick isn't an option for them (side note, iirc popular Linux distros like Ubuntu already ship with it? I never installed it but I still use it daily)

Though if you really wanted a web-based converter towebp looks fine to me. I double checked and couple find any mention of limits and they do the conversion locally which is nice

binaryoverload: I agree CLI is preferable, I just wanted an easy way for people to convert the images without having to install anything 

I'll switch to towebp 👍

PR #375, opened 2025-02-23
Added a NNID warnings and removed outdated info on Juxt

### LINKED ISSUE #374: [Enhancement]: Add a warning for deleting NNIDs
### Checked Existing

- [x] I have checked the repository for duplicate issues.

Unfortunately, people follow the wrong guides or just randomly decide that deleting the NNID for whatever reason will fix any of the issues with Pretendo, but it won't. The "022-5432" or "022-2613" error code problem will happen regardless if you have the NNID or not.

We have seen people unfortunately deleting the NNID and then asking why they can't play their games anymore (on Wii U) or can't do specific features on the console because of the NNID being removed, so assuming people don't even have backups to begin with.

This has happened a lot of times during these almost 11 months since the shutdown of Nintendo Network and a lot of misinformation is still spreading online how to setup Pretendo. Adding this warning will probably do nothing in the short-term but in long-term people will be more aware overtime, reducing potential misinformation being spread in the future.

We do care about people not losing their data when installing homebrew projects, and so this should be clear that no one ever should advice anymore to delete the NNID just to install Pretendo (if it's because they just want to format the console from zero, have anyone's else NNID then it's absolutely valid doing it, also in very rare issues about the NNID). For the purpose of setting up Pretendo, "deleting/removing the NNID" shouldn't be not recommended at all and avoided to prevent more inconvienences in the future.

_No response_

piplupness:
Resolves #374 

### Changes:
Added a disclaimer about not removing your NNID and fixed outdated info (Juxtaposition is public for almost a year now and rverse is currently dead)


MiniBioOne:
i think it'd be better to say "your NNID" instead of "the NNID"

docs/en_US/install/3ds.md
```diff
@@ -1,5 +1,10 @@
 # 3DS/2DS Family
 
+<div class="tip red">
+	<strong>!!! DISCLAIMER !!!
+	DON'T REMOVE THE NNID TO SET UP PRETENDO NETWORK, IT'S NOT NECESSARY AND BY DOING SO YOU WILL LOSE BADGES, REDOWNLOADABLE GAMES, THEMES AND POKÉMON BANK DATA!</strong>
+</div>
```
mrjvs, line 6:
```suggestion
<div class="tip red">
	<strong>CAUTION:</strong>
	DON'T REMOVE YOUR NNID TO SET UP PRETENDO NETWORK, IT'S NOT NECESSARY AND BY DOING SO YOU WILL LOSE BADGES, REDOWNLOADABLE GAMES, THEMES AND POKÉMON BANK DATA!
</div>
```

docs/en_US/install/wiiu.md
```diff
@@ -1,5 +1,10 @@
 # Wii U
 
+<div class="tip red">
+	<strong>!!! DISCLAIMER !!!
+	DON'T REMOVE THE NNID TO SET UP PRETENDO NETWORK, IT'S NOT NECESSARY AND BY DOING SO YOU WILL LOSE ALL YOUR PREVIOUSLY PURCHASED GAMES AND UNABLE TO UPDATE THEM!</strong>
+</div>
```
mrjvs, line 6:
```suggestion
<div class="tip red">
	<strong>CAUTION:</strong>
	DON'T REMOVE YOUR NNID TO SET UP PRETENDO NETWORK, IT'S NOT NECESSARY AND BY DOING SO YOU WILL LOSE ALL YOUR PREVIOUSLY PURCHASED GAMES AND WILL BE UNABLE TO UPDATE THEM!</strong>
</div>
```

mrjvs (changes requested):

PR #377, opened 2025-03-17
Add privacy and cookie policies

### LINKED ISSUE #296: Define a clear privacy policy
_Apologies if this is the wrong repository_
As I was planning on registering a pretendo account I noticed there was no defined privacy policy or terms of service to inform me of how my data will be used if I register. Looking at other projects (Wiimifi) a clear privacy policy is defined in order to inform users of how their data is used and their right to erasure. As pretendo becomes more popular I believe this is something that must be addressed.

jonbarrow:
Partially resolves #296

### Changes:

Adds a clear privacy and cookie policy, as per #296. However I do not feel that this fully closes the issue, as we are still lacking a few things like an EULA. That being said, I feel that this PR alone is good enough for now, and we should get it merged right away once we are happy with the verbage (I have done my best, but I am no legal expert)

Some notes:

1. Yes I am aware that this looks somewhat rushed, and that's because it is. I am aware that the policies pages reuse the blog pages layouts. I just want to get something out the door here, since #296 has been open for FAR too long
2. We do not have a solidified EULA yet, and adding one is going to require several other changes to places like the account server. Documents such as those are out of scope for this PR (the cookie policy only made it in here because it was part of the privacy policy creators I used)
3. A link has been added to the website footer to link to the privacy policy, which itself has a link to the cookie policy. I believe this to be enough, as like I said I want to get this out the door as quickly as possible and anything else would require larger changes which we can address later. My concern is just getting the pages on the website in some form right now
4. I did not write these policies entirely by hand, so please keep that in mind when reviewing. See below for details
 
These terms are almost entirely a mashup of generated/pre-made documents from both https://termly.io/ and https://www.termsfeed.com/, which were then merged together and modified to the best of our ability to fit our specific needs. However we are not legal experts, and there is quite a lot of legalese here, and possibly broad policies with regards to collection/retention/sharing which we do NOT necessarily want/need. Our goal is to collect as LITTLE information about users as possible, and to never sell or directly share that data with any 3rd parties besides those required for operation, such as Cloudflare and Stripe. I have done my best to remove all the broad policies that would allow us to sell user data, as we do NOT want to do that, but I am no legal expert and may have missed things. In those cases, we are more than happy to revise/remove the problematic areas to clarify our goals/intentions

If anyone is unhappy with any wording or policies, do feel free to voice that. There's a large chance that I wasn't even aware something was in here, as most of this was made by just clicking some buttons in a form initially

To reiterate, we do not collect any data that is not required to run our services and we don't sell data or anything like that. There are cases where some 3rd parties collect your data, but they are required for us to operate such as Cloudflare. If there is anything in the document which seems to go against that, please do let us known


src/routes/terms.js
```diff
@@ -0,0 +1,35 @@
...
+router.get('/:slug', async (request, response, next) => {
+	const renderData = {
+		layout: 'blog-opengraph'
+	};
+
+	const termName = request.params.slug;
+
+	let rawTerm;
+	try {
+		rawTerm = await fs.readFile(path.join('terms', `${termName}.md`), 'utf-8');
```
mrjvs, line 19: path traversal?

jonbarrow: Good catch, this is actually in the blog posts too

jonbarrow:
These documents are basically identical to the ones that people like Billy already said looked fine internally. I just tweaked them a tiny bit to add more instances of stuff like “we don’t sell your data”

So if there’s no objections by morning (in around 8-ish hours) I’ll probably just move forward with the merge

PR #393, opened 2025-04-13
feat: add navbar + i8n

gitlimes:
Added navbar and basic locale support. 
This should probably get merged so that all other contributions can make use the locale support.


src/layouts/default.vue
mrjvs, line 1: There are linting errors in this file. Run `npm run lint:fix` to fix them.

Although if you use vscode, the formatting should already be preconfigured for you, we may have to look into why it isn't doing this for you.

src/locales/el_GR.json
```diff
@@ -184,7 +184,7 @@
...
 			"Ήξερες ότι το \"Ξέρεις ότι η Ash, νά' ναι καλά η ψυχούλα, κάνει όλη μέρα UwU\" είναι ο ευγενικός τρόπος να πείς \"Η Ash κάνει UwU συνέχεια και είναι περίεργο και χαζό και μακάρι να μην το έκανε;\"",
-			"Το πρώτο βίντεω στο κανάλι μου!! περίμενα να κάνω βίντεος για πολι καιρό αλλά το λαπτοπ μου ετρεθε πολυ κακά και δεν μπορουσα να τρεχω το fraps, skype κι minecraft ολα μαζι. αλλα αυτο τελιωσε! με λιγο βοηθεια απο το δασκαλο πληροφορικισ μου το λαπτ0π μου τρεχι πολυ καλυτερα και μπορο να καταγράψω τορα! ευχομαι να το απολαυσετε και αν ναι, παρ@καλο καντε λαεκ και σαμπσκραϊμπ!"
+			"Το πρώτο βίντεω στο κανάλι μου!! περίμενα να κάνω βίντεος για πολι καιρό αλλά το λαπτοπ μου ετρεθε πολυ κακά και δεν μπορουσα να τρεχω το fraps, skype κι minecraft ολα μαζι. αλλα αυτο τελιωσε! με λιγο βοηθεια απο το δασκαλο πληροφορικισ μου το λαπτ0π μου τρεχι πολυ καλυτερα και μπορο να καταγράψω τορα! ευχομαι να το απολαυσετε και αν ναι, παρ{'@'}καλο καντε λαεκ και σαμπσκραϊμπ!"
```
mrjvs, line 187: Does vue-i18n treat the `@` as special syntax?

src/components/Navbar/Navbar.vue
```diff
@@ -0,0 +1,1259 @@
...
+header nav a.donate button.dropdown-button::after {
+	content: "";
+	display: inline-block;
+	align-items: center;
+	width: 16px;
+	height: 8px;
+	bottom: 0;
+	margin-left: 0.3rem;
+	background: no-repeat center url("/assets/images/down-arrow.svg");
+	filter: brightness(0) invert(78%) sepia(2%) saturate(5488%) hue-rotate(197deg) brightness(88%) contrast(93%);
```
mrjvs, line 844: The image is missing, let's also copy it into the project (should be fine to drop it into the [public directory](https://nuxt.com/docs/guide/directory-structure/public))

src/components/Navbar/Navbar.vue
```diff
@@ -0,0 +1,1259 @@
...
+}
+
+header .user-widget .button.logout {
+	background: var(--bg-shade-3);
+	color: var(--text-shade-3);
+}
+
+@media screen and (max-width: 900px) {
+	header {
+		position: relative;
```
mrjvs, line 1159: I must admit I haven't tried it before. But I don't think media queries work in a scoped style block?
Simple solution is to just make another `<style>` (not scoped) and put all the media queries in there.

src/components/Navbar/Navbar.vue
```diff
@@ -0,0 +1,1259 @@
...
+}
+
+header div.dropdown .top a:hover::after {
+	content: "";
+	position: absolute;
+	top: 0;
+	right: 24px;
+	width: 24px;
+	height: 100%;
+	background: no-repeat center url("/assets/images/arrow-right.svg");
```
mrjvs, line 936: Another image

src/components/Navbar/Navbar.vue
```diff
@@ -0,0 +1,1259 @@
...
+	}
+}
+</script>
+
+<template>
+  <header class="transparent">
+    <div class="left-section">
+      <button
+        id="mobile-button"
+        class="dropdown-button"
```
mrjvs, line 22: Let's remove the `id` attribute from the buttons, there's no use for it now (plus it can cause problems later, browsers don't like multiple ids).

There are a few more ids in this file, I wont add a review comment for all of them cuz its a bit spammy.

You can keep the `clippath` ids as is. It does cause problems but those existed in the original codebase so I don't think its a problem. (For now)

src/components/Navbar/Navbar.vue
```diff
@@ -0,0 +1,1259 @@
+<script setup>
```
mrjvs, line 1: This is something I forgot to mention, but we should use typescript in our component files. You can enable it by simply adding `lang="ts"` to the script tag.

With that enabled you can also add a type to the first parameter of `handleDropdownButton`

mrjvs (changes requested):
Sorry about the amount of review comments, it's a bit more than I anticipated.

For anybody following along. We decided to basically copy paste the existing site into before we make it good.

src/layouts/default.vue
gitlimes: I am, that is weird.

src/locales/el_GR.json
gitlimes: It does!

https://vue-i18n.intlify.dev/guide/essentials/syntax.html#special-characters

src/components/Navbar/Navbar.vue
gitlimes: i am a fool

src/locales/el_GR.json
mrjvs: Annoying, but makes sense

mrjvs (approved):

PR #394, opened 2025-04-14
feat: footer

gitlimes:
footer? i hardly know her


src/components/Footer/Footer.vue
```diff
@@ -0,0 +1,440 @@
...
+// @ts-ignore
+const totalBandwidthQuotes = messages.value['en-US'].footer.bandwidthRaccoonQuotes.length;
+
+function bandwidthClickHandler() {
+	if (bandwidthQuoteIndex.value + 1 >= totalBandwidthQuotes) {
+		bandwidthQuoteIndex.value = 0;
+	} else {
+		bandwidthQuoteIndex.value++;
+	}
+}
```
mrjvs, line 16: Sorry, I couldn't resist trying to fix the ts-ignore and accidentally overcomplicated it.

(all you need to add to this is using `currentQuote` in the template below)

```suggestion
const { tm, t } = useI18n();
const year = new Date().getFullYear();

const bandwidthQuoteIndex = ref<number>(-1);
const quotes = computed(() => tm<string>('footer.bandwidthRaccoonQuotes') as string[]);
const totalBandwidthQuotes = computed(() => quotes.value.length);
const currentQuote = computed(() => t(`footer.bandwidthRaccoonQuotes[${Math.max(0, bandwidthQuoteIndex.value)}]`));

watch([totalBandwidthQuotes], () => {
	bandwidthQuoteIndex.value = -1;
});

function bandwidthClickHandler() {
	if (bandwidthQuoteIndex.value + 1 >= totalBandwidthQuotes.value) {
		bandwidthQuoteIndex.value = 0;
	} else {
		bandwidthQuoteIndex.value++;
	}
}
```

src/components/Footer/Footer.vue
```diff
@@ -0,0 +1,440 @@
...
+	border-color: var(--bg-shade-3) transparent;
+	border-width: 12px 12px 0;
+	bottom: -9px;
+	right: 60px;
+	margin-left: -10px;
+}
+</style>
+
+<style>
+/* So many !important??? there must be an easier way */
```
mrjvs, line 367: you can add more specificity. for example if we add `#root` in the default.vue. we can then overwrite all styles by just doing `#root footer`. then !important wouldn't be needed since our style is more specific than the vue one.

(the vue one is `footer[data-v-????]`)

mrjvs: We can also add `lang="scss"` to the style tag, then we can use nesting:
```css
@media screen and (max-width: 900px) {
  #root {
    footer {}
    footer div {}
    footer div.discord-server-card-wrapper {}
    // etc
  }
}
```

mrjvs (changes requested):

src/components/Footer/Footer.vue
gitlimes: oh my god. ok thanks jvs

mrjvs (approved):

PR #395, opened 2025-04-15
Nuxt refactor homepage

gitlimes:
homepage :)


src/components/Navbar/Navbar.vue
```diff
@@ -12,13 +12,23 @@ function handleDropdownButton(dropdown: boolean | string) {
 		openDropdown.value = dropdown;
 	}
 }
+
+const scrollY = ref<number>(0);
+
+onMounted(() => {
+	// eslint-disable-next-line no-undef -- we're in on mounted so it should be fine
```
binaryoverload, line 19: My bad, I didn't configure the globals

You can use the example from https://eslint.vuejs.org/user-guide/#configuration-eslint-config-js, set the globals as they show. You don't need the rules section or source type though

src/pages/index.vue
```diff
@@ -1,5 +1,1567 @@
...
+            href="https://github.com/PretendoNetwork/"
+            class="do-button-margin"
+            target="_blank"
+            aria-label="Open Github"
+          >
+            <button
+              class="button secondary github icon-btn"
+              focusable="false"
+            >
+              <svg
```
binaryoverload, line 98: I think we should probably pull the SVG files out into their own files and import them - that way this file isn't cluttered with long SVG definitions

binaryoverload (changes requested):
Looking good! Just a couple comments from me

gitlimes:
Gotcha! Gonna work on it this evening. Btw @binaryoverload should I be requesting a review from you too from now onwards?

binaryoverload:
@gitlimes you can keep the reviews just to JVS  - I'll just pop in every now and then :) 

JVS and I are hive mind anyway, so you're getting both of us either way 😛

src/pages/index.vue
gitlimes: It might also just be a good idea to directly use the phosphoricons vue module, since most of the icons come from there

mrjvs: We have `nuxt/icon` installed Which allows you to do: `<Icon name="ph:address-book-fill" />`.

It supports everything from iconify, which includes phosphor icons: https://icon-sets.iconify.design/ph/

gitlimes: gotcha, that's good

gitlimes:
i am a dumbass who is sleep deprived and asked you for a review anyways. whoops

binaryoverload (approved):
Thanks for implementing those changes - looks good from my perspective

PR #407, opened 2025-07-27
Delete Text for Removing NNID on Wii U Install Guide

Joshua8600:
Changes:

1. Delete Text for Removing NNID on Wii U Install Guide

Reasoning:
The act of removing an NNID is unnecessary at best, harmful at worst.

Having that text there contradicts the "Caution" at top of the page, and also might lead people to believe that it is okay to do that.

This PR removes that text, and copies the one from the 3DS page (along with a grammar edit).


Joshua8600:
@mrjvs Here's your ping

PR #425, opened 2026-03-16
add passwordrules with pattern and add birthdate to locales

### LINKED ISSUE #164: Add password rules to web signup
Apple's Password Rules are used by iCloud Password and other password manager extensions to ensure autofilled passwords matched the required schema; using the information from my Wii U showing the available characters (this _should_ be checked on 3DS just in case, though I doubt Nintendo would perform the same issue themselves...), as well as the following information from Nintendo:

> The password you choose should be between 6 and 16 characters long.
Passwords must use a combination of characters from two of the following categories: letters, numbers and punctuation.
You cannot use the name you chose for your Nintendo Network ID as your password.
The same character cannot be used three or more times in a row.

The following password rules should be added to the <input> HTML tag:
```
passwordrules="minlength: 6; maxlength: 16; required: lower; required: upper; required: digit; allowed: [!&quot;$%&'()*+,./:;<=>?@[\^_`{|}~]];"
```
As well as  `maxlength="16"` to ensure only a PNID password of 16 characters can be input. A regex may also be a good idea for non-password-manager environments as they can also ensure only accepted characters are input. 10,000 passwords using this rule have been attached generated using [Apple's website](https://developer.apple.com/password-rules/), so we can validate the rule.

[generated-passwords.txt](https://github.com/PretendoNetwork/website/files/9611933/generated-passwords.txt)

To use the rule yourself on Apple's website, enter the following:
```
required: lower;
required: upper;
required: digit;
allowed: [~!@$%^&*()-+=_`:;\|"'<>.,?/{}[]]
```

LelouchFR:
Resolves #164 

### Changes:

changes in the `register.handlebars` template file to get `passwordrules` attribute with the content of #164.

update the form to unhard-code birthdate to add it to locales like the other values so it can be translated using weblate

- [X] What I'm implementing was an [approved issue](../issues?q=is%3Aopen+is%3Aissue+label%3Aapproved).


jonbarrow (approved):
LGTM

PR #450, opened 2026-08-20
feat: account editing on web

limesdotpink:
### Changes:
- added account editing on web

**REQUIRES https://github.com/PretendoNetwork/account/pull/351**


server/utils/useGrpc.ts
```diff
@@ -53,6 +53,6 @@ export function useLegacyApiGrpcWithToken(event: H3Event, token: string): Client
 	return getGrpcClient(event, APIDefinition, token);
 }
 
-export function useApiGrpcWithToken(event: H3Event, token: string): Client<APIDefinition> {
-	return getGrpcClient(event, APIDefinition, token);
+export function useApiGrpcWithToken(event: H3Event, token: string): Client<ApiServiceDefinition> {
+	return getGrpcClient(event, ApiServiceDefinition, token);
```
mrjvs, line 57: Just gonna mention it here, but this did have the opportunity to break stuff. but I checked the implementations of the v1 vs v2 and there are no differences. Everything should still work.

src/pages/account/index.vue
```diff
@@ -253,54 +330,319 @@ useHead({
...
-              </AlertDialog.Title>
-              <AlertDialog.Description class="modal-caption">
-                <p>
-                  {{
-                    $t("account.settings.settingCards.no_edit_from_dashboard")
-                  }}
-                </p>
-              </AlertDialog.Description>
+          </Dialog.Trigger>
+          <Dialog.Portal :to="dialogContainer ?? undefined">
```
mrjvs: Lets make a new component: `AccountInfoEditModal`. It's large enough to isolate the code into it's own component.

You can use `defineEmits()` to create a `submit` event which can be used to refetch the new info on this page. The actual saving can be done inside the modal component.

You can probably get away with only putting the `Dialog.Content` in the component. So the trigger and the rest can stay the same.
Alternatively you can make a slot for the trigger.

src/pages/account/index.vue
```diff
@@ -60,18 +68,68 @@ const { data: connections, refresh: refreshConnections } = await useApiFetch(
...
-const editModalOpen = ref(false);
-const selectedServerEnv = ref<'dev' | 'test' | 'prod' | undefined>(profile.value?.serverAccessLevel);
+const profileEditModalOpen = ref(false);
+const emailEditModalOpen = ref(false);
+const selectedServerEnv = ref<'dev' | 'test' | 'prod' | undefined>(
+	profile.value?.serverAccessLevel
+);
+
+const newBirthday = shallowRef(parseDate(profile.value?.birthday || ''));
+const newCountry = ref(((profile.value?.region || 0) >>> 24) & 0xff);
```
mrjvs: bitwise is quite hard to read, I'd recommend placing the fancy bitwise operations in helper methods. Then the method name can describe what the bitwise is supposed to do. e.g. `extractCountryFromRegion(region: number)`

src/pages/account/index.vue
```diff
@@ -253,54 +330,319 @@ useHead({
...
+                        region: newRegion,
+                        timezone: newTimezone,
+                      })
+                  "
+                >
+                  <Loader v-if="isLoadingUpdateUserData" />
+                  <span v-else>{{
+                    $t("modals.confirm")
+                  }}</span>
+                </Dialog.Close>
```
mrjvs: Let's make this a standard button, then the spinner is actually shown to the user before it closes.

src/pages/account/index.vue
```diff
@@ -253,54 +330,319 @@ useHead({
...
+                    </Select.Content>
+                  </Select.Root>
+                </fieldset>
+              </div>
+
               <div class="modal-button-wrapper">
-                <AlertDialog.Cancel class="cancel">
-                  {{ $t("modals.close") }}
-                </AlertDialog.Cancel>
+                <Dialog.Close class="cancel">
```
mrjvs: Make sure to add `disabled` when it's currently saving. we dont want the user to close it and not see it fail. (or worse, reopen and edit again during a save)

mrjvs (changes requested):

src/pages/account/index.vue
limesdotpink: Good call, yeah.

Ended up making a slot for the trigger, so it's cleaner.

src/pages/account/index.vue
limesdotpink: Added `regionIdToCountryId(region: number)` and `countryIdToUndefinedRegionId(region: number)`

src/pages/account/index.vue
limesdotpink: Done that, thx ;)

src/pages/account/index.vue
limesdotpink: Hadn't thought of that. Done :t-rex:

src/pages/account/index.vue
```diff
@@ -1,14 +1,11 @@
 <script setup lang="ts">
 /* eslint-disable vue/no-v-html -- locale files still have raw html */
-import { watchImmediate, watchOnce } from '@vueuse/core';
-import { AlertDialog, Label, Dialog, Select, DatePicker } from 'reka-ui/namespaced';
-import { parseDate } from '@internationalized/date';
+import { watchImmediate } from '@vueuse/core';
+import { AlertDialog, Dialog } from 'reka-ui/namespaced';
+import AccountInfoEditModal from '~/components/AccountInfoEditModal.vue';
```
mrjvs: components dont need to be imported. Nuxt has autoimports for everything in components directory

src/pages/account/index.vue
```diff
@@ -1,14 +1,11 @@
...
 import {
-	getLocalizedRegionTimezones,
 	getLocalizedTimezoneString,
-	regionIdToLocalizedNames,
-	getLocalizedCountryList,
-	getLocalizedRegionList
+	regionIdToLocalizedNames
```
mrjvs, line 7: import looks unused?

mrjvs (changes requested):
Couple of small things.

After the account PR is merged, i'm gonna do a bit of testing as a final review and then it can be merged.

limesdotpink:
Thanks, will work on these tomorrow evening

src/pages/account/index.vue
limesdotpink: I'm not sure what this is referring to? This made me notice that I'm importing `Dialog` from reka when it's not actually necessary, but it's not what you've highlighted here... am I missing something?

package.json
mrjvs, line 1: Please update the docker-compose.yml with the new dev-branch image so it supports account editing.

(I cant comment on docker-compose.yml cuz it wasnt changed, so im using package.json)

src/components/AccountInfoEditModal.vue
```diff
@@ -0,0 +1,742 @@
...
+			toasts.publish({
+				type: 'error',
+				text: err.message
+			});
+		}
+	});
+</script>
+
+<template>
+  <Dialog.Root v-model:open="open">
```
mrjvs, line 86: While it's saving you can still close the modal by clicking on the backdrop. Maybe its good to prevent that from happening as well? (cuz the buttons were disabled properly)

mrjvs (changes requested):
After some testing, two more small things:

limesdotpink:
On it!

mrjvs (approved):
lgbt

limesdotpink:
okay then, do i merge and deploy?

PR #454, opened 2026-08-26
feat: implement email editing/verification endpoint

limesdotpink:
### Changes:
- Change email
- Verify email endpoint for emails to point to
- Resend verification email

**REQUIRES** https://github.com/PretendoNetwork/account/pull/355


server/api/account/update-email.patch.ts
```diff
@@ -0,0 +1,38 @@
+import { ClientError } from 'nice-grpc';
+import { EmailUpdateSchema } from '~~/shared/api-types';
+import type { ApiErrorCodes } from '~~/shared/errors';
+
+const bucket = createRatelimitBucket({
+	id: 'update-email',
+	points: 3,
```
mrjvs: Let's give it a couple more points, it also counts invalid emails as attempts.
Lets say 10 points?

server/api/account/verify-email.post.ts
```diff
@@ -0,0 +1,36 @@
+import { ClientError } from 'nice-grpc';
+import { EmailVerifySchema } from '~~/shared/api-types';
+import type { ApiErrorCodes } from '~~/shared/errors';
+
+const bucket = createRatelimitBucket({
+	id: 'verify-email',
+	points: 3,
```
mrjvs: We can be more relaxed on this. lets do 20 points.
Scraping it is practically impossible anyway

mrjvs (changes requested):
I'll give it a proper test later when account is merged and I have more time.

But after looking through all the code this is all I can find.
(except for the localisation keys, but you already know about that)

server/api/account/update-email.patch.ts
limesdotpink: right, hadn't considered that.

server/api/account/verify-email.post.ts
limesdotpink: okie!

src/components/UpdateEmailModal.vue
```diff
@@ -0,0 +1,184 @@
...
+                disabled:
+                  isLoadingUpdateEmail
+              }"
+              :disabled="isLoadingUpdateEmail"
+              @click="executeUpdateEmail"
+            >
+              {{
+                $t("account.settings.unverifiedEmailModal.resend")
+              }}
+            </button>
```
mrjvs, line 155: This doesn't have a visible loading state, Try adding the loading component we had. I did something similar at the discord integration.

src/components/UpdateEmailModal.vue
```diff
@@ -0,0 +1,184 @@
...
+            {{ $t('account.settings.unverifiedEmailModal.lostEmail') }}
+            <button
+              :class="{
+                'update-trigger':
+                  true,
+                disabled:
+                  isLoadingUpdateEmail
+              }"
+              :disabled="isLoadingUpdateEmail"
+              @click="executeUpdateEmail"
```
mrjvs: Resending verification email does not work.
This is the error: `/api.v2.ApiService/UpdateEmail INVALID_ARGUMENT: INVALID_ARGUMENT: Must provide new email address`
The input email is an empty string.

src/components/UpdateEmailModal.vue
```diff
@@ -0,0 +1,184 @@
...
+          </button>
+        </div>
+      </Dialog.Content>
+    </Dialog.Portal><Dialog.Trigger class="update-trigger">
+      {{ $t('account.settings.update') }}
+    </Dialog.Trigger>
+  </Dialog.Root>
+
+  <Dialog.Root
+    v-else
```
mrjvs: This `v-else` means you can't update your email it's not verified. Which puts my test account in an interesting spot: it's unverified with an unfunctional email and I'm not allowed to change it's not verified yet.

src/pages/account/verify-email.client.vue
```diff
@@ -0,0 +1,85 @@
...
+	},
+	onError(error) {
+		const err = getApiError(error);
+		errorMessage.value = err.message;
+	}
+});
+
+await callOnce(async () => {
+	await executeVerifyEmail();
+});
```
mrjvs: If we want the users to see the loading spinner then this shouldn't be awaited.

mrjvs (changes requested):
I went through the tests and found a couple things.
I haven't been able to test if the account server actually sends emails, but everything else I've managed to test using some DB manipulation.

I like the UI, it turned out very nice!

package.json
mrjvs, line 1: Forgot one review comment: Please add the new account service image to the docker compose so the dev setup keeps working.

mrjvs (changes requested):

limesdotpink:
Thank you!

Should be good to merge now

mrjvs (approved):

mrjvs:
LGTM

PR #461, opened 2026-08-31
fix(#460): #fragment links not working

### LINKED ISSUE #460: [Bug]: `#fragment` links seem to no longer work
### Checked Existing

- [x] I have checked the repository for duplicate issues.

### What happened?

Clicking on `#fragment` links updates the address bar in the browser, but doesn't jump to that section on the page

### What did you expect to happen?

The links should jump to the fragment

### Steps to reproduce?

1. Visit a page with fragments and links to them, like https://pretendo.network/docs/network-dumps
2. Click a fragment link
3. The address bar updates to show the fragment but your position on the page does not change

### Other relevant information. (OPTIONAL)

Visiting the page with the fragment already in the URL does jump to that section, it's only when clicking the links does it fail

I've tried this on both the latest versions of MS Edge and Google Chrome on MacOS and had the issue on both browsers, so I don't think it's a browser bug

limesdotpink:
Resolves #460

### Changes:
- fixed the bug (duh)

- [X] What I'm implementing was an [approved issue](../issues?q=is%3Aopen+is%3Aissue+label%3Aapproved).

### DIFF
--- a/src/pages/docs/[...slug].vue
+++ b/src/pages/docs/[...slug].vue
@@ -1,5 +1,7 @@
 <script setup lang="ts">
-const slug = (useRoute().params.slug as string[]).join('/');
+const { path, params } = useRoute();
+const slug = (params.slug as string[]).join('/');
+
 const { data: doc } = await useAsyncData(`docs-${slug}`, () => {
 	return queryCollection('docs').path(`/docs/${slug}`).first();
 });
@@ -7,6 +9,25 @@ const { data: doc } = await useAsyncData(`docs-${slug}`, () => {
 if (!doc.value) {
 	throw createError({ statusCode: 404, statusMessage: 'Page Not Found' });
 }
+
+function fixFragmentLinks(e: MouseEvent) {
+	const targetEl = e.target as HTMLAnchorElement;
+	const to = targetEl?.getAttribute('href');
+
+	if (targetEl?.tagName !== 'A' || !to?.startsWith('#')) {
+		return;
+	}
+
+	// nuxt content just drops dashes at the start of the fragment, so we do the same
+	const sanitizedTo = to.replace(/^#-*/, '');
+
+	// a fragment cannot start with a number, so nuxt content prefixes it with an underscore.
+	const newPath = /^[0-9]/.test(sanitizedTo) ? `${path}#_${sanitizedTo}` : `${path}#${sanitizedTo}`;
+
+	e.preventDefault();
+	return navigateTo(newPath, { external: true });
+}
+
 customSeoMeta({
 	subsection: 'docs',
 	title: doc.value.title,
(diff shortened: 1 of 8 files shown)


src/pages/docs/[...slug].vue
```diff
@@ -23,6 +44,7 @@ definePageMeta({
     <ContentRenderer
       v-if="doc"
       :value="doc"
+      @click="fixFragmentLinks"
```
Comment (mrjvs) [inline src/pages/docs/[...slug].vue]: Instead of relying on a (sortof) global click handler. I'd rather that we replace markdown links with our own rendering:

```vue
<ContentRenderer
  :components="{
    a: 'MyLinkComponent'
  }"
/>
```

Then we can add the custom scrollTo behaviour in there.
Use a `<NuxtLink>` in your custom component, then if it's a fragment. just do a `document.getElementById(id)?.scrollTo()`

src/pages/docs/[...slug].vue
```diff
@@ -1,12 +1,33 @@
...
+
+	if (targetEl?.tagName !== 'A' || !to?.startsWith('#')) {
+		return;
+	}
+
+	// nuxt content just drops dashes at the start of the fragment, so we do the same
+	const sanitizedTo = to.replace(/^#-*/, '');
+
+	// a fragment cannot start with a number, so nuxt content prefixes it with an underscore.
+	const newPath = /^[0-9]/.test(sanitizedTo) ? `${path}#_${sanitizedTo}` : `${path}#${sanitizedTo}`;
```
Comment (mrjvs) [inline src/pages/docs/[...slug].vue]: Instead of recalculating the fragments to be like nuxt-content. I'd rather we update the markdown docs so it doesn't contain the wrong fragment links.

With this syntax you can specify a custom ID on the headers:
```
# Hello world {#hello-world}
```
See [here](https://content.nuxt.com/docs/files/markdown#attributes)

src/pages/docs/[...slug].vue
```diff
@@ -1,12 +1,33 @@
...
+
+	e.preventDefault();
+	return navigateTo(newPath, { external: true });
```
Comment (mrjvs) [inline src/pages/docs/[...slug].vue]: We can use `document.getElementById(url.slice(1))?.scrollTo()` to prevent a page reload. Just make sure to check if it's a fragment link first.

mrjvs (changes requested):

src/pages/docs/[...slug].vue
Comment (limesdotpink) [reply src/pages/docs/[...slug].vue]: The headers were never the issue, the links to said headers were. I was trying to avoid editing the markdown, but if we're gonna do that then we can just change the links themselves:

```diff
- [3DS (HokakuCTR)](#3ds-hokakuctr)
+ [3DS (HokakuCTR)](#_3ds-hokakuctr)
```

mrjvs (approved):
LGTM

